PR #24, opened 2021-08-22
Migration: SwiftMailer to Symfony/Mailer

Bl00D4NGEL:
refs https://github.com/rectorphp/rector/issues/6654

This is just the beginning of the migration and a few more things need to be done like replacing
```php
        $this->swift->setSender($address, $name);
```
with
```php
        is_array($address)
            ? $this->email->sender(...$address)
            : $this->email->sender(new Address($address, (string) $name));
```

etc. but I figured this might be too big for one PR and should therefore be split up into more PRs.

I'm also not sure about the SetList (if we should already "release it" while it isn't complete yet)

### DIFF
--- a/tests/Rector/MethodCall/SwiftCreateMessageToNewEmailRector/Fixture/fixture.php.inc
+++ b/tests/Rector/MethodCall/SwiftCreateMessageToNewEmailRector/Fixture/fixture.php.inc
@@ -2,9 +2,13 @@
 
 namespace Rector\Symfony\Tests\Rector\MethodCall\SwiftCreateMessageToNewEmailRector\Fixture;
 
+use Swift_Mailer;
+
 class Fixture
 {
-    public function __construct(\Swift_Mailer $swift)
+    private $swift;
+
+    public function __construct(Swift_Mailer $swift)
     {
         $this->swift = $swift;
     }
@@ -21,9 +25,13 @@ class Fixture
 
 namespace Rector\Symfony\Tests\Rector\MethodCall\SwiftCreateMessageToNewEmailRector\Fixture;
 
+use Swift_Mailer;
+
 class Fixture
 {
-    public function __construct(\Swift_Mailer $swift)
+    private $swift;
+
+    public function __construct(Swift_Mailer $swift)
     {
         $this->swift = $swift;
     }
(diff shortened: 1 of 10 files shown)


tests/Rector/MethodCall/SwiftCreateMessageToNewEmailRector/Fixture/fixture.php.inc
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace Rector\Symfony\Tests\Rector\MethodCall\SwiftCreateMessageToNewEmailRector\Fixture;
+
+class Fixture
+{
+    public function __construct(\Swift_Mailer $swift)
```
Bl00D4NGEL, line 7: I was hoping that rector would be able to detect that `$this->swift` will be of type `\Swift_Mailer` in order to be a little safer when refactoring but this doesn't seem to be the case, did I miss something for that?

TomasVotruba: In some cases PHPStan needs to have stubs in `/stubs` directory.

TomasVotruba:
> I'm also not sure about the SetList (if we should already "release it" while it isn't complete yet)

We usually release is ASAP, so we can test it. Upgrade is never complete, so go for it :+1:

TomasVotruba:
Looks good, lets kit it of :smiley:

samsonasik:
@TomasVotruba should it has new release first before release rector-src ?

TomasVotruba:
@samsonasik I think we can wait a bit here, since this set will need much more work to be able to handle  migration correctly.

Bl00D4NGEL:
I guess we should start collecting what needs to be done so what other people can pick up the work and help us, right?
For example, looking at the laraval PR we can see that a few is_array($xxx) checks need to be added, a few classes need to be replaced etc. Adding this to a list like it was done in https://github.com/rectorphp/rector/issues/3127 or https://github.com/rectorphp/rector/issues/6436

TomasVotruba:
Mimicing https://github.com/laravel/framework/pull/38481/files would be the easiest start :+1:

PR #27, opened 2021-08-27
[Symfony] Skip no arg on PropertyAccessorCreationBooleanToFlagsRector

samsonasik:
No arg on `PropertyAccessor` new instance creation means use default value.


samsonasik:
@TomasVotruba can we have new rector-symfony release for this patch so it will be included in next rector release?

PR #65, opened 2021-11-14
Avoid ContainerGetToConstructorInjectionRector on TestCase classes

### LINKED ISSUE #64: ContainerGetToConstructorInjectionRector should ignore PHPUnit or KernelTestCase classes
When having `ContainerGetToConstructorInjectionRector` activated it tries to refractor the PHPUnit KernelTestCase classes to constructor. But as this are not service this should be avoided.

I tried to contribute it but I found not yet a way how I can check in https://github.com/rectorphp/rector-symfony/blob/6f0f3d16e82f48e26690b9382dd015ad92bbf391/src/Rector/MethodCall/ContainerGetToConstructorInjectionRector.php#L77
if the `$node` is inside a PHPUnit TestCase class / KernelTestCase class.

alexander-schranz:
fixes #64

### DIFF
--- a/src/Rector/MethodCall/ContainerGetToConstructorInjectionRector.php
+++ b/src/Rector/MethodCall/ContainerGetToConstructorInjectionRector.php
@@ -87,6 +87,12 @@ CODE_SAMPLE
             return null;
         }
 
+        $classLike = $this->betterNodeFinder->findParentType($node, \PhpParser\Node\Stmt\ClassLike::class);
+        $className = $classLike->namespacedName->toString();
+        if (in_array('PHPUnit\Framework\TestCase', class_parents($className))) {
+            return null;
+        }
+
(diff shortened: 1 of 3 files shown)


TomasVotruba:
Thank you 👍 

We'll need a falling test case to be sure it won't be broken in the future

alexander-schranz:
@TomasVotruba can you help me here. The `skip_test_case.php.inc` already exist but seems not be executed. All things I change there are just ignored and I'm not sure why.

TomasVotruba:
In cases test passes without change, there is bug elsewhere.
Could you create a reproducible repository on Github with minimal failing code?

alexander-schranz:
@TomasVotruba have a reproducable for my current open PRs: https://github.com/alexander-schranz/rector-reproducable-repository

TomasVotruba:
Thanks, I can see it now :+1:

PR #66, opened 2021-11-14
Avoid ContainerGetToConstructorInjectionRector on TestCase classes

TomasVotruba:
Closes https://github.com/rectorphp/rector-symfony/pull/65

Closes https://github.com/rectorphp/rector-symfony/issues/64


TomasVotruba:
@alexander-schranz Your solution was correct :+1: Just in this case it's better to use PHPStan scope instead of class_parents. The PHPStan failing CI should suggest that.

PR #67, opened 2021-11-19
StringToArrayArgumentProcessRector breaks Process instance method calls

mrardon:
This is a failing test to show the issue. Basically I think it is overzealous with the instance methods of the Process class.

The Process class API for `setWorkingDirectory` is
```
public function setWorkingDirectory(string $cwd)
```

The StringToArrayArgumentProcessRector changes the proper `string` argument to an `array` argument when it should not.


samsonasik:
I cherry-picked your commit at PR https://github.com/rectorphp/rector-symfony/pull/70

PR #70, opened 2021-11-22
[Symfony] Do not convert to array Process->setWorkingDirectory()

samsonasik:
Closes #67

### DIFF
--- a/src/Rector/New_/StringToArrayArgumentProcessRector.php
+++ b/src/Rector/New_/StringToArrayArgumentProcessRector.php
@@ -13,16 +13,11 @@ use PhpParser\Node\Expr\BinaryOp\Concat;
 use PhpParser\Node\Expr\FuncCall;
 use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\New_;
-use PhpParser\Node\Identifier;
 use PhpParser\Node\Scalar\String_;
-use PhpParser\Node\Stmt\ClassMethod;
-use PHPStan\Analyser\Scope;
 use PHPStan\Type\ObjectType;
 use PHPStan\Type\StringType;
-use Rector\Core\PhpParser\AstResolver;
 use Rector\Core\PhpParser\NodeTransformer;
 use Rector\Core\Rector\AbstractRector;
-use Rector\NodeTypeResolver\Node\AttributeKey;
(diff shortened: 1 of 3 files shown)


samsonasik:

src/Rector/New_/StringToArrayArgumentProcessRector.php
```diff
@@ -96,6 +102,22 @@ private function processArgumentPosition(New_|MethodCall $node, int $argumentPos
             return null;
         }
 
+        if ($node instanceof MethodCall) {
```
TomasVotruba, line 105: AST resolver is quite heavy to handle this case.
There should be check for allowed method names, that can be converted.

samsonasik: implemented https://github.com/rectorphp/rector-symfony/pull/70/commits/67323a8e664834a510efe4098eddeb92a744a299

samsonasik:

src/Rector/New_/StringToArrayArgumentProcessRector.php
```diff
@@ -34,9 +29,13 @@
  */
 final class StringToArrayArgumentProcessRector extends AbstractRector
 {
+    /**
+     * @var string[]
+     */
+    private const ALLOWED_PROCESS_METHOD_CALLS = ['setWorkingDirectory'];
```
TomasVotruba, line 35:
```suggestion
    private const EXCLUDED_PROCESS_METHOD_CALLS = ['setWorkingDirectory'];
```

TomasVotruba: I'm not sure why there are `MethodCall` in the first place.
Maybe it would be better to `INCLUDED_PROCESS_METHOD_CALLS` explicitly and name the methods, that can be used in this rule.

samsonasik: updated

mrardon: I think this may be the wrong approach. Pretty much any method in the Process that uses a string variable is being converted to array arguments. It seems that only the constructor should be changed from a string to array. Not every method call in the class. Maybe I'm missing something. Take a look at the other public methods `addOutput` and `addErrorOutput` they have the same issue as my original #67

PR #71, opened 2021-11-22
Overzealous string to array conversion

mrardon:
Relates to #67 and #70

I went through to check that #70 covered all the cases and it looks like at least 2 methods are missed with this approach. I have added those methods here.


TomasVotruba:
Thank you 😊

PR #87, opened 2021-12-08
[Dep] Remove stringy as replaced with symfony/string

samsonasik:
@TomasVotruba Looking at other rector packages, eg:

https://github.com/rectorphp/rector-nette/commit/1e3b3ab4e1248623b020ce9c8e744ad89e942e74

The `danielstjules/stringy` was removed, I think it should be removed as well.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,6 @@
     "require": {
         "php": ">=8.1",
         "ext-xml": "*",
-        "danielstjules/stringy": "^3.1",
         "symfony/string": "^6.0"
     },
     "require-dev": {


samsonasik:
All checks have passed 🎉 @TomasVotruba I think it is ready. @sabbelasichon typo3-rector may need update first to remove stringy before it completely gone.

TomasVotruba:
Yes, thank you. I forgot to verify the PR I made

samsonasik:
I added tag rector-symfony 0.11.49 https://github.com/rectorphp/rector-symfony/releases/tag/0.11.49 for it as there is already new release for typo3-rector.

@TomasVotruba it seems new symplify packages releases is needed to make it completely gone from vendor.

PR #108, opened 2022-02-14
[Stub][Test] Fix Tests error for AbstractController stub usage on using latest symfony 6

samsonasik:
ref https://github.com/rectorphp/rector-src/pull/1812#issuecomment-1038330946 on using latest symfony 6

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     "require-dev": {
         "phpunit/phpunit": "^9.5",
         "phpstan/phpstan": "^1.3",
-        "rector/rector-src": "dev-main",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -11,7 +11,7 @@
     "require-dev": {
         "phpunit/phpunit": "^9.5",
         "phpstan/phpstan": "^1.3",
-        "rector/rector-src": "dev-main",
+        "rector/rector-src": "dev-main#cb2822cbfe3dc71ab2a72305590014c406c3639d",
```
samsonasik, line 14: temporary per-PR https://github.com/rectorphp/rector-src/pull/1812 usage

TomasVotruba:
:partying_face:

Issue #112, opened 2022-02-28
Symfony Userinterface::getUsername => getUserIdentifier only for Symfony >= 6.0


mkrauser:
Symfony changes the name getUsername in the UserInterface to getUserIdentifier with Symfony 6.0.

Using the new name works starting at Symfony 5.3, but since the UserInterface is not changed yet, renaming the method in Symfony 5.3 and 5.4 will break the code, since the User-Class implementing the Interface is missing getUsername()

This Rule should be moved to the Symfony 6.0 set.

see 
https://github.com/symfony/symfony/blob/5.3/src/Symfony/Component/Security/Core/User/UserInterface.php
https://github.com/rectorphp/rector-symfony/blob/8aa9a6b20bdbc9859cdd82f03aa975572cc35eb0/config/sets/symfony/symfony53.php#L41

mkrauser:
Sure, here it is

samsonasik:
Fixed at https://github.com/rectorphp/rector-symfony/pull/113

PR #113, opened 2022-02-28
renaming getUsername to getUserIdentifier should only be done for Symfony 6.0

mkrauser:
see https://github.com/rectorphp/rector-symfony/issues/112

Renaming it in Symfony 5.3 and 5.4 would break the code, since the getUsername() is missing from the respective User-Class, but still required by the UserInterface


TomasVotruba:
Thank you :+1: 

Could you also add a note about your findings, why it should be in Symfony 6? Just to keep juggling back and forth without context

mkrauser:
Just did, is a link to my issue enough?

TomasVotruba:
Yes, thanks 👍

PR #115, opened 2022-03-04
Controller-Actions can only be public, non static functions

### LINKED ISSUE #114: ControllerMethodAnalyzer::isAction may return true for private/protected functions
the issue is located [here](https://github.com/rectorphp/rector-symfony/blob/5c1486a2b4840eebb51ebec70c79684313dc0c6a/src/Bridge/NodeAnalyzer/ControllerMethodAnalyzer.php#L36)

Even if the name of the containing class ends with "Controller", imho a private or protected method cannot be action. Or am I missing something here?

Also, the Action-Suffix is not enough evidence either. It's use is discouraged. So the only thing we are left with, is to check if the method is inside a controller and see if it is public and not static.

Checking for an existing route to that controller is not an option, because there may be actions without a route. This is often used for internal sub-requests. 

I'll submit a PR. 

Thank you!

mkrauser:
fixes #114

### DIFF
--- a/src/Bridge/NodeAnalyzer/ControllerMethodAnalyzer.php
+++ b/src/Bridge/NodeAnalyzer/ControllerMethodAnalyzer.php
@@ -7,13 +7,13 @@ namespace Rector\Symfony\Bridge\NodeAnalyzer;
 use PhpParser\Node;
 use PhpParser\Node\Stmt\ClassMethod;
 use PHPStan\Analyser\Scope;
-use Rector\NodeCollector\ScopeResolver\ParentClassScopeResolver;
 use Rector\NodeTypeResolver\Node\AttributeKey;
+use Rector\Symfony\TypeAnalyzer\ControllerAnalyzer;
 
 final class ControllerMethodAnalyzer
 {
     public function __construct(
-        private readonly ParentClassScopeResolver $parentClassScopeResolver
+        private readonly ControllerAnalyzer $controllerAnalyzer
     ) {
     }
 
@@ -31,15 +31,10 @@ final class ControllerMethodAnalyzer
             return false;
         }
 
-        $parentClassName = (string) $this->parentClassScopeResolver->resolveParentClassName($scope);
-        if (\str_ends_with($parentClassName, 'Controller')) {
-            return true;
+        if (! $this->controllerAnalyzer->isInsideController($node)) {
+            return false;
         }
 
-        if (\str_ends_with((string) $node->name, 'Action')) {
-            return true;
-        }
-
-        return $node->isPublic();
+        return $node->isPublic() && ! $node->isStatic();
     }
 }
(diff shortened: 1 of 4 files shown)


mkrauser:
Regarding the failing test: 

Since the class in the [fixture](https://github.com/rectorphp/rector-symfony/blob/main/tests/Rector/ClassMethod/ActionSuffixRemoverRector/Fixture/fixture2.php.inc) is not a Subclass of AbstractController, the rector should indeed skip this class. 

I know that a Controller-Class is not required to inherit from AbstractRector, but this is usually the case. If we don't want to rely on this we can check for the Controller-Suffix in the class-name.

Let me know, what you think!

TomasVotruba:
> Since the class in the [fixture](https://github.com/rectorphp/rector-symfony/blob/main/tests/Rector/ClassMethod/ActionSuffixRemoverRector/Fixture/fixture2.php.inc?rgh-link-date=2022-03-04T13%3A05%3A00Z) is not a Subclass of AbstractController, the rector should indeed skip this class.

Agreed :+1: 

Please also add test case to cover your original issue.

mkrauser:
> Please also add test case to cover your original issue.

Since my issue was with a custom rule, I wrote a low-level Test-Case for ControllerMethodAnalyzer, I hope that's OK.

TomasVotruba:
>  I wrote a low-level Test-Case for ControllerMethodAnalyzer, I hope that's OK.

Oh, I thought this rule is broken.
Okay, but better use service container test and avoid any mocking. It's pain to maintain and does not any value.

mkrauser:
> Okay, but better use service container test and avoid any mocking. It's pain to maintain and does not any value.

Thanks for the tip. I'll try that tomorrow

tests/Rector/ClassMethod/ActionSuffixRemoverRector/Fixture/fixture2.php.inc
```diff
@@ -21,7 +21,7 @@ use Rector\Symfony\Tests\Rector\Source\AbstractSymfonyController;
 
 class AppController2 extends AbstractSymfonyController
 {
-    public function index()
+    public function indexAction()
```
TomasVotruba, line 24: Also, it would be easier to duplicate one or converted fixture and add `static` there.

mkrauser: > Also, it would be easier to duplicate one or converted fixture and add `static` there.

You mean, that adding for fixtures for this test is enough, since it indirectly tests ControllerMethodAnalyzer?
I did just refactor my TestCase from the PR use the container, but I still have to create the Nodes manually.

TomasVotruba: > You mean, that adding for fixtures for this test is enough, since it indirectly tests ControllerMethodAnalyzer

Exactly! 👍

TomasVotruba:
Thank you :+1:

PR #117, opened 2022-03-06
[Symfony] Handle property exists on ContainerGetToConstructorInjectionRector

samsonasik:
Fixes https://github.com/rectorphp/rector-symfony/issues/111

PR #118, opened 2022-03-06
fix infinite loop on ContainerGetToConstructorInjectionRector

samsonasik:
Continue of https://github.com/rectorphp/rector-symfony/pull/117, given the following code cause infinite loop:

```php

final class SomeTranslatorExists3 extends AbstractController
{
    private SomeTranslator $someTranslator;
    private SomeTranslator $someTranslator2;

    public function __construct(SomeTranslator $someTranslator, SomeTranslator $someTranslator2)
    {
        $this->someTranslator = $someTranslator;
        $this->someTranslator2 = $someTranslator2;
    }

    protected function execute()
    {
        $someService = $this->getContainer()->get('translator');

        $someService = $this->getContainer()->get('translator')->translateSomething();
    }
}
```

This PR fix it.


TomasVotruba (commented):
👍

PR #120, opened 2022-03-08
[Symfony] Only Increment property name when different object on ContainerGetToConstructorInjectionRector

samsonasik:
Make `ContainerGetToConstructorInjectionRector` only increment naming for new property when name is equal but different object, use existing when exists.

- [x] Same name generation, object type is equal? -> use as is
- [x] Different name generation, object type is equal? -> use as is
- [x] Same name generation, object type is different ? -> increment naming


samsonasik:
implemented 🎉 , rector-nette's test need to be reverted if this PR merged https://github.com/rectorphp/rector-nette/pull/57

samsonasik:

samsonasik:
I am merging it ;)

PR #121, opened 2022-03-08
Fixing rectify run

samsonasik:
Current github action seems always pending on rectify on rector-symfony, this PR add unused property on purpose to run rectify...


samsonasik:
it seems rectify github action still pending, even changing to `GITHUB_TOKEN`

Screen Shot 2022-03-08 at 17 07 36

I am closing it for now.

Issue #123, opened 2022-03-11
Rector failed on default Symfony config run if PHPUnit is not installed.


bocharsky-bw:
See "Class PHPUnit\Framework\TestCase not found." error in the output:
```
$ ./vendor/bin/rector process src/


 [WARNING] Your project requires min PHP version "7.4".

           1 Rector rules defined in your configuration require higher PHP version and will not run,
           to avoid breaking your codebase, use -vvv for detailed info.


 ! [NOTE] Do you want to run them? Make "require" > "php" in `composer.json` higher,
 !        or add "Option::PHP_VERSION_FEATURES" parameter to your `rector.php`.


In RuntimeReflectionProvider.php line 185:

  Class PHPUnit\Framework\TestCase not found.


process [-n|--dry-run] [-a|--autoload-file AUTOLOAD-FILE] [--no-progress-bar] [--no-diffs] [--output-format OUTPUT-FORMAT] [--memory-limit MEMORY-LIMIT] [--clear-cache] [--port PORT] [--identifier IDENTIFIER] [--] [<source>...]
```

Just for the reference, my `rector.php` config is:
```
return static function (ContainerConfigurator $containerConfigurator): void {
    // region Symfony Container
    $parameters = $containerConfigurator->parameters();
    $parameters->set(
        Option::SYMFONY_CONTAINER_XML_PATH_PARAMETER,
        __DIR__ . '/var/cache/dev/App_KernelDevDebugContainer.xml'
    );
    // endregion

    $containerConfigurator->import(SymfonyLevelSetList::UP_TO_SYMFONY_54);
    $containerConfigurator->import(SymfonySetList::SYMFONY_CODE_QUALITY);
    $containerConfigurator->import(SymfonySetList::SYMFONY_CONSTRUCTOR_INJECTION);
};
```

Installing PHPUnit is a workaround for this.

TomasVotruba:
Hi, if some rule is working with PHPUnit, it will require it's dependency, so this behavior is expected.

Could you share the demo link that shows this error? https://getrector.org/demo

bocharsky-bw:
I'm not sure if it's possible to do with that https://getrector.org/demo feature as I'm talking about the whole Symfony project... I was trying to apply the default Symfony config there but it shows this error:

> Failed to read "/var/www/var/demo/15y7lp7hy5ju1otiox9a/var/cache/dev/App_Ke
rnelDevDebugContainer.xml" file: ""

[...]

PR #125, opened 2022-03-12
[Symfony] Skip SimplifyWebTestCaseAssertionsRector on no PHPUnit\Framework\TestCase class

samsonasik:
Fixes https://github.com/rectorphp/rector-symfony/issues/123

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -38,6 +38,7 @@ return static function (ContainerConfigurator $containerConfigurator): void {
             'Twig_*',
             'Swift_*',
             'Doctrine\*',
+            'PHPUnit\Framework\TestCase',
         ]);
 
     $containerConfigurator->import(LevelSetList::UP_TO_PHP_81);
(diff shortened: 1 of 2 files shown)


samsonasik:
@bocharsky-bw this should fix https://github.com/rectorphp/rector-symfony/issues/123

samsonasik:

I verified run it in https://github.com/bocharsky-bw/no-phpunit-rector-reproducer repository.

TomasVotruba:
I'm not sure if this is correct, as the webtest case methods will not be changed. 
Instead we should provide somehow test case stub. 

Even better, the correct solution is adding the dependency, as Rector/PHPStan can work only with existing classes:
https://github.com/rectorphp/rector/issues/6531#issuecomment-870586660

samsonasik:
@TomasVotruba Ok, I created alternative PR at https://github.com/rectorphp/rector-src/pull/1924

PR #132, opened 2022-03-15
Add set to migrate JMS annotations to attributes

Jean85:
This should solve https://github.com/rectorphp/rector/issues/7052

I don't know if this needs any test.

### DIFF
--- a/config/sets/jms/annotations-to-attributes.php
+++ b/config/sets/jms/annotations-to-attributes.php
@@ -1,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+use Rector\Php80\Rector\Class_\AnnotationToAttributeRector;
+use Rector\Php80\ValueObject\AnnotationToAttribute;
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+
+return static function (ContainerConfigurator $containerConfigurator): void {
+    $services = $containerConfigurator->services();
+    $services->set(AnnotationToAttributeRector::class)
+        ->configure([
+            new AnnotationToAttribute('JMS\Serializer\Annotation\Accessor'),
+            new AnnotationToAttribute('JMS\Serializer\Annotation\AccessorOrder'),
+            new AnnotationToAttribute('JMS\Serializer\Annotation\AccessType'),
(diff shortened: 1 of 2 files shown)


samsonasik (approved):

Jean85:
I just noticed that these rules don't work in all situations. One common case is mixed named and non-named arguments, like:
```php
 * @JMS\AccessorOrder("custom", custom = {"id", "email"})
```
after applying Rector it goes:
```php
#[JMS\AccessorOrder(custom: ['id', 'email'])]
```
loosing the `name` argument.

Is this an issue on Rector's side or a custom behavior from JMS?

Maybe we can ask @goetas..

config/sets/jms/annotations-to-attributes.php
```diff
@@ -0,0 +1,45 @@
...
+use Rector\Php80\ValueObject\AnnotationToAttribute;
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+
+return static function (ContainerConfigurator $containerConfigurator): void {
+    $services = $containerConfigurator->services();
+    $services->set(AnnotationToAttributeRector::class)
+        ->configure([
+            new AnnotationToAttribute('JMS\Serializer\Annotation\Accessor'),
+            new AnnotationToAttribute('JMS\Serializer\Annotation\AccessorOrder'),
+            new AnnotationToAttribute('JMS\Serializer\Annotation\AccessType'),
```
TomasVotruba, line 15: Could you link the responsible PR in jms repository? Just for future context

TomasVotruba:
> I just noticed that these rules don't work in all situations. One common case is mixed named and non-named arguments, like:

That's interesting bug on Rector's side. I'll check it in standalone PR

config/sets/jms/annotations-to-attributes.php
Jean85: There are multiple ones:
 * https://github.com/schmittjoh/serializer/pull/1320
 * https://github.com/schmittjoh/serializer/pull/1332
 * https://github.com/schmittjoh/serializer/pull/1337

TomasVotruba: I mean in the set as a comment. It is very helpful to have them available in case of edge cases

PR #138, opened 2022-04-03
Add ParamTypeFromRouteRequiredRegexRector

TomasVotruba:


samsonasik:
@TomasVotruba the github action after rectify on rector-* packages seems pending, I tried using `GITHUB_TOKEN` but seems still pending, ref https://github.com/rectorphp/rector-symfony/pull/121, need close-re-open to restart build

TomasVotruba:
@samsonasik Thanks for heads-up. I've just finished, let' see how CI handles it :)

PR #143, opened 2022-04-08
Failing test to demostrate desired behavior for extending a class still needing parents constructor.

ccovey:
Based on conversation from twitter.

I expect this is not actually a problem for rector to attempt to solve actually so we may just close it but i'll give the case below and we can discuss. 

We have a legacy symfony app that uses base controllers grabbing services out of the container extensively for certain behaviors used across controllers throughout our application. This was the norm for a long time obviously but now with the smaller service locator container passed into the `AbstractController` I need to be able to move these services to constructor injection. Rector handles anything the concrete controllers are using but the base controllers dependencies are ignored. 

The other option would be to use a SubscriberService method to provide services to add to the service locator, the issue is symfony requires these to be class names which all of our dependencies are not (things like service classes with multiple definitions) so thats a no go mostly as well. 

I have a feeling the "proper" way will be to suck it up and refactor out most of the "base controllers" functionality out and do dependency injection more pointedly but figured a failing test with the outline above can at least spark a conversation and worst case is you close this and I can get to work on the refactor a different way.

### DIFF
--- a/tests/Rector/MethodCall/GetToConstructorInjectionRector/Fixture/fixure3.php.inc
+++ b/tests/Rector/MethodCall/GetToConstructorInjectionRector/Fixture/fixure3.php.inc
@@ -1,0 +1,42 @@
+<?php
+
+namespace Rector\Symfony\Tests\Rector\MethodCall\GetToConstructorInjectionRector\Fixture;
+
+use Symfony\Bundle\FrameworkBundle\Controller\BaseController;
+
+class ClassWithNamedService extends BaseController
+{
+    public function render()
+    {
+        $someService = $this->get('some_service');
+
+        $this->renderTwig([
+            'posts' => $this->get('some_service')->callMe()
+        ]);
+    }
+}
+
+?>
+-----
+<?php
+
+namespace Rector\Symfony\Tests\Rector\MethodCall\GetToConstructorInjectionRector\Fixture;
+
+use Symfony\Bundle\FrameworkBundle\Controller\BaseController;
+
+class ClassWithNamedService extends BaseController
+{
+    public function __construct(private \stdClass $stdClass, private \stdClass $formFactory)
+    {
+    }
+    public function render()
+    {
+        $someService = $this->stdClass;
+
+        $this->renderTwig([
+            'posts' => $this->stdClass->callMe()
+        ]);
+    }
+}
+
+?>
--- a/stubs/Symfony/Bundle/FrameworkBundle/Controller/BaseController.php
+++ b/stubs/Symfony/Bundle/FrameworkBundle/Controller/BaseController.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Symfony\Bundle\FrameworkBundle\Controller;
+
+use Symfony\Component\Form\FormInterface;
+use Symfony\Component\HttpFoundation\Response;
+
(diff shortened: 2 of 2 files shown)


tests/Rector/MethodCall/GetToConstructorInjectionRector/Fixture/fixure3.php.inc
```diff
@@ -0,0 +1,42 @@
...
+-----
+<?php
+
+namespace Rector\Symfony\Tests\Rector\MethodCall\GetToConstructorInjectionRector\Fixture;
+
+use Symfony\Bundle\FrameworkBundle\Controller\BaseController;
+
+class ClassWithNamedService extends BaseController
+{
+    public function __construct(private \stdClass $stdClass, private \stdClass $formFactory)
```
TomasVotruba, line 29: The form factory cannot be here, as it's handled by the Symfony controller itself:

```suggestion
    public function __construct(private \stdClass $stdClass)
```

TomasVotruba:
From a short code view, it seem you're mixing 2 refactorings:

* moving from named services in controllers that extend Symfony controller to constructor injection 
* moving away from Symfony controller to pure constructor injection

These should be 2 separated refactoring. In my experience, the latter seems clean and academic, but it bring huge maintenance cost and kills the DX that native Symfony controller brings. I tried to refactor away from Symfony, but after few months of writing own framework, I realized one abstract class that every Symfony developer understands is just fine and much better than my "cool" complex code :)

TomasVotruba:
> The other option would be to use a SubscriberService method to provide services to add to the service locator, the issue is symfony requires these to be class names which all of our dependencies are not (things like service classes with multiple definitions) so thats a no go mostly as well.

This depends on Symfony version and I'd have to see more context of your specific project to answer propperly. Yet Rector itself also uses abstract class that all entry points inherit from - the `AbstractRector`.

It needs couple services itself, to avoid repeating in every child Rector rule. We could do the parent injecting and have huge consturctors in every Rector... or we can use the Required annotation/attribute - https://symfony.com/doc/current/service_container/autowiring.html#autowiring-other-methods-e-g-setters-and-public-typed-properties:
https://github.com/rectorphp/rector-src/blob/4fb405ed3a6f5287c3b54800dc57d33a7cde2df0/src/Rector/AbstractRector.php#L139-L164
There you can autowire by type, not by a name.

Nette has similar mechanism to abstract from parent constructors via inject annotation/attribute: https://doc.nette.org/en/dependency-injection/passing-dependencies#toc-passing-by-an-inject-method

The rule is to use those only in `abstract` classes of course. There are even PHPStan rules to handle that.

ccovey:
So I took a bit of the advice here and just slammed through the refactors separately. First I got rid of anything from the base controllers that was unused or only used in 1 or 2 controllers and set up service subscribers for the rest. Then from there I was able to run the above rule on the concrete controllers (I used the exclude configurations to leave the base controllers alone during this step.) It was a bit of manual work to get through it simply because I needed to audit what was actually used.

This was a good exercise for me either way and got to play with a bit of rector also. I appreciate the nudge in the right direction here!

TomasVotruba:
Thanks for reaching out with full story. I'm glad you made it and happily surprised it took you such a short time :clap: 

When I come across similar stuggles with Rector in legacy projects I deal with, I always try to step up and think "what would human do?". I often realize the problem can be split into 2-3 smaller chunks that are easier to handle.

Congrats again, and if you had any more questions, feel free to reach out here.

PR #147, opened 2022-04-11
Add missing annotations to symfony52-validator-attributes set

guillaumesmo:
Support for all theses attributes was added in Symfony 5.2, together with the already present Groups attribute
see https://github.com/symfony/symfony/pull/38525

there is only one missing Serializer annotation, but it was added in Symfony 5.4, so I left that one out
https://github.com/symfony/symfony/pull/38525


TomasVotruba:
Thank you 👍

PR #151, opened 2022-04-20
 [DX] Update to use Method Call from RectorConfig on rector configs 

samsonasik:


tests/Rector/Class_/EventListenerToEventSubscriberRector/config/config_with_parameter.php
```diff
@@ -8,11 +8,8 @@
 
 return static function (RectorConfig $rectorConfig): void {
     $rectorConfig->import(__DIR__ . '/../../../../../config/config.php');
-
     $parameters = $rectorConfig->parameters();
     // wtf: all test have to be in single file due to autoloading race-condigition and container creating issue of fixture
     $parameters->set(Option::SYMFONY_CONTAINER_XML_PATH_PARAMETER, __DIR__ . '/listener_services.xml');
```
samsonasik, line 13: @TomasVotruba `$parameters` definition can't be removed as needs on `$parameters->set(Option::SYMFONY_CONTAINER_XML_PATH_PARAMETER)` so I change it manually.

TomasVotruba:
Ref https://github.com/rectorphp/rector-src/pull/2109

PR #155, opened 2022-04-24
Split SimplifyWebTestCaseAssertionsRector into 3 particular rules

TomasVotruba:


config/config.php
```diff
@@ -17,6 +17,7 @@
     $services->load('Rector\\Symfony\\', __DIR__ . '/../src')
         ->exclude([__DIR__ . '/../src/{Rector,ValueObject}']);
 
-    $services->set(RenameClassNonPhpRector::class);
+    $rectorConfig->rule(RenameClassNonPhpRector::class);
```
samsonasik, line 20: this seems should be kept to use `$services->set()` to avoid error:

```                                                                                                    
 [ERROR] Expected an instance of this class or to this class among his parents                                          
         Rector\Core\Contract\Rector\RectorInterface. Got: string in                                                    
         /Users/samsonasik/www/rector-symfony/config/config.php (which is being imported from                           
         "/Users/samsonasik/www/rector-symfony/rector.php").      
```

ref https://github.com/rectorphp/rector-laravel/pull/43

TomasVotruba: The `RenameClassNonPhpRector` class should be always present and autoloaded. It might have been an old vendor dependency.

PR #158, opened 2022-04-26
Move Renaming of UserProviderInterface::loadUserByUsername() from Symfony 5.3 to Symfony 6.0

dritter:
Hi there!

UserProviderInterface::loadUserByUsername() is deprected in Symfony 5.3, but it is still defined in the interface and will be removed in the next Symfony Major Release (6.0). Prior to Symfony 6.0 rector would create broken code.

Compare UserProviderInterface in [Symfony 5.3](https://github.com/symfony/security-core/blob/5.3/User/UserProviderInterface.php#L68) and [Symfony 6.0](https://github.com/symfony/security-core/blob/6.0/User/UserProviderInterface.php#L63).

Cheers!

### DIFF
--- a/config/sets/symfony/symfony53.php
+++ b/config/sets/symfony/symfony53.php
@@ -44,11 +44,6 @@ return static function (RectorConfig $rectorConfig): void {
                 'getUserIdentifier',
             ),
             new MethodCallRename(
-                'Symfony\Component\Security\Core\User\UserProviderInterface',
-                'loadUserByUsername',
-                'loadUserByIdentifier',
-            ),
-            new MethodCallRename(
                 'Symfony\Component\Security\Core\Exception\UsernameNotFoundException',
                 'getUsername',
                 'getUserIdentifier'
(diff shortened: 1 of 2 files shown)


stephanvierkant:
What's the problem with renaming it while using Symfony 5.3?

dritter:
Sorry. I should have explained it better..
So, in Symfony 6.0 the `UserProviderInterface::loadUserByUsername()` was renamed to `loadUserByIdentifier()`. For a short period of time, that change was made on Symfony 5.3, but was quickly [rolled back](https://github.com/symfony/security-core/commit/22714af1f701937a0a0bd3e3ec2a761baed3f2d0) (BC break). But this change is still applied in Rector. Long story short: If you run this rule against your < Symfony 6 code base, you are left with broken code because your code does not fulfill the UserProviderInterface any more.

stephanvierkant:
Ah, that makes sense! +1 for this PR

TomasVotruba:
Thank you :+1:

PR #161, opened 2022-05-03
Remove obsolete ArgumentAdder

### LINKED ISSUE #160: SymfonyLevelSetList::UP_TO_SYMFONY_42 BrowserKit crawler method update parsing throws exception
When bumping the SymfonyLevelSetList of a legacy application to `UP_TO_SYMFONY_42`, parsing the code resulted in the following exception:

    [ERROR] Could not process "tests/Controller/Test.php" file, due to:                                  
             "System error: "Invalid identifier name "[]""

I've chased that down to the following minimal code that will trigger the failure above:

```php
/** @test */
public function formTest(): void
{
    $client = static::createClient();

    $crawler = $client->request('GET', '/api/path/here');

    $form = $crawler->selectButton('Button Text Here')->form();

    $client->submit($form);

}
```

I've traced that code _triggering_ the behavior as far as https://github.com/rectorphp/rector-symfony/blob/main/config/sets/symfony/symfony42.php#L66

With that knowledge, I modified the code above to the following:

```php
/** @test */
public function formTest(): void
{
    $client = static::createClient();

    $crawler = $client->request('GET', '/api/path/here');

    $form = $crawler->selectButton('Button Text Here')->form();

    $client->submit($form, []);
}
```
Doing so resolves the exception and allows rector to complete analysis of all code; however, it proposes a change incompatible with the function:

```diff
-        $client->submit($form, []);
+        $client->submit($form, [], []);
```

With `symfony/browser-kit` 4.4.37 vendored, the submit function has the following signature:

```php
public function submit(Form $form, array $values = [])
```

Given the above, I can work around this for the time being, but wanted to document this issue I encountered.

jrmcpeek:
Upstream projected reverted the backward-compatibility breaking change.

Fixes #160.

### DIFF
--- a/config/sets/symfony/symfony42.php
+++ b/config/sets/symfony/symfony42.php
@@ -61,17 +61,6 @@ return static function (RectorConfig $rectorConfig): void {
 
     $rectorConfig
         ->ruleWithConfiguration(ArgumentAdderRector::class, [
-            // https://github.com/symfony/symfony/commit/fa2063efe43109aea093d6fbfc12d675dba82146
-            // https://github.com/symfony/symfony/commit/e3aa90f852f69040be19da3d8729cdf02d238ec7
-            new ArgumentAdder(
-                'Symfony\Component\BrowserKit\Client',
-                'submit',
-                2,
-                'serverParameters',
-                [],
-                null,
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Thank you :+1:

PR #163, opened 2022-05-05
Add CommandPropertyToAttributeRector

### LINKED ISSUE #159: Feature idea: Add the Symfony\Component\Console\Attribute\AsCommand attribute
As of Symfony 6.1, I've got this notification in my logs:

> Since symfony/console 6.1: Relying on the static property "$defaultDescription" for setting a command description is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "App\Command\NotificationCommand" class instead.

I'd the great if Rector can refactor commands with `static $defaultName` and `static $defaultDescription`.

stephanvierkant:
My first PR that adds a Rector, so happy to get some feedback.

Fixes #159

### DIFF
--- a/src/Rector/Class_/CommandPropertyToAttributeRector.php
+++ b/src/Rector/Class_/CommandPropertyToAttributeRector.php
@@ -1,0 +1,146 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Symfony\Rector\Class_;
+
+use PhpParser\Node;
+use PhpParser\Node\Arg;
+use PhpParser\Node\AttributeGroup;
+use PhpParser\Node\Scalar\String_;
+use PhpParser\Node\Stmt\Class_;
+use PhpParser\Node\Stmt\Property;
+use PHPStan\Reflection\ReflectionProvider;
+use PHPStan\Type\ObjectType;
+use Rector\Core\Rector\AbstractRector;
+use Rector\Core\ValueObject\PhpVersionFeature;
+use Rector\Php80\NodeAnalyzer\PhpAttributeAnalyzer;
+use Rector\PhpAttribute\Printer\PhpAttributeGroupFactory;
+use Rector\VersionBonding\Contract\MinPhpVersionInterface;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+/**
+ * @see https://symfony.com/doc/current/console.html#registering-the-command
+ *
+ * @see \Rector\Symfony\Tests\Rector\Class_\CommandPropertyToAttributeRector\CommandPropertyToAttributeRectorTest
+ */
+final class CommandPropertyToAttributeRector extends AbstractRector implements MinPhpVersionInterface
+{
+    private const ATTRIBUTE = 'Symfony\Component\Console\Attribute\AsCommand';
+
+    public function __construct(
+        private readonly PhpAttributeGroupFactory $phpAttributeGroupFactory,
+        private readonly PhpAttributeAnalyzer $phpAttributeAnalyzer,
+        private readonly ReflectionProvider $reflectionProvider
+    ) {
+    }
+
+    public function provideMinPhpVersion(): int
+    {
+        return PhpVersionFeature::ATTRIBUTES;
+    }
+
+    public function getRuleDefinition(): RuleDefinition
+    {
+        return new RuleDefinition(
+            'Add Symfony\Component\Console\Attribute\AsCommand to Symfony Commands and remove the deprecated properties',
+            [
+                new CodeSample(<<<'CODE_SAMPLE'
+class SunshineCommand extends \Symfony\Component\Console\Command\Command
+{
+    /** @var string|null */
+    public static $defaultName = 'sunshine';
+}
+CODE_SAMPLE
+                , <<<'CODE_SAMPLE'
+#[\Symfony\Component\Console\Attribute\AsCommand('sunshine')]
+class SunshineCommand extends \Symfony\Component\Console\Command\Command
+{
+}
+CODE_SAMPLE),
+
+            ]
+        );
+    }
+
+    /**
+     * @return array<class-string<Node>>
+     */
+    public function getNodeTypes(): array
+    {
+        return [Class_::class];
+    }
+
+    /**
+     * @param Class_ $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if (! $this->isObjectType($node, new ObjectType('Symfony\\Component\\Console\\Command\\Command'))) {
+            return null;
+        }
+
+        if (! $this->reflectionProvider->hasClass(self::ATTRIBUTE)) {
+            return null;
+        }
+
+        if ($this->phpAttributeAnalyzer->hasPhpAttribute($node, self::ATTRIBUTE)) {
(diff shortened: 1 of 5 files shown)


src/Rector/Class_/CommandPropertyToAttributeRector.php
```diff
@@ -0,0 +1,114 @@
...
+     * @var \Rector\Php80\NodeAnalyzer\PhpAttributeAnalyzer
+     */
+    private $phpAttributeAnalyzer;
+    /**
+     * @readonly
+     * @var \Rector\PhpAttribute\Printer\PhpAttributeGroupFactory
+     */
+    private $phpAttributeGroupFactory;
+
+    public function __construct(\Rector\PhpAttribute\Printer\PhpAttributeGroupFactory $phpAttributeGroupFactory, \Rector\Php80\NodeAnalyzer\PhpAttributeAnalyzer $phpAttributeAnalyzer)
```
samsonasik, line 29: property promotion and readonly can be used here. You can run:

```
vendor/bin/rector process src/Rector/Class_/CommandPropertyToAttributeRector.php
```

src/Rector/Class_/CommandPropertyToAttributeRector.php
```diff
@@ -0,0 +1,120 @@
...
+ * @see \Rector\Symfony\Tests\Rector\Class_\CommandPropertyToAttributeRector\CommandPropertyToAttributeRectorTest
+ */
+final class CommandPropertyToAttributeRector extends AbstractRector implements MinPhpVersionInterface
+{
+    private const ATTRIBUTE = 'Symfony\Component\Console\Attribute\AsCommand';
+
+    public function __construct(
+        /**
+         * @readonly
+         */
```
samsonasik, line 31: docblock readonly no longer needed

```suggestion
```

src/Rector/Class_/CommandPropertyToAttributeRector.php
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * @param Class_ $node
+     */
+    public function refactor(Node $node) : ?Node
+    {
+        if (!$this->isObjectType($node, new ObjectType('Symfony\\Component\\Console\\Command\\Command'))) {
+            return null;
+        }
+
+        if (!class_exists(self::ATTRIBUTE)) {
```
samsonasik, line 78: it should use reflectionProvider instead:

```suggestion
        if (! $this->reflectionProvider->hasClass(self::ATTRIBUTE)) {
```

You can inject dependency to `__construct()`:

```php
private readonly \PHPStan\Reflection\ReflectionProvider $reflectionProvider
```

src/Rector/Class_/CommandPropertyToAttributeRector.php
```diff
@@ -0,0 +1,120 @@
...
+            return null;
+        }
+
+        if ($this->phpAttributeAnalyzer->hasPhpAttribute($node, self::ATTRIBUTE)) {
+            return null;
+        }
+
+        $defaultName = null;
+        $defaultDescription = null;
+
```
samsonasik, line 88: when `count($node->props) > 1`, it should be skipped:

```php
if (count($node->props) > 1) {
    return null;
}
```

src/Rector/Class_/CommandPropertyToAttributeRector.php
```diff
@@ -0,0 +1,120 @@
...
+        $property = $node->getProperty('defaultName');
+        if ($property !== null) {
```
samsonasik, line 90:
```suggestion
        if ($property instanceof \PhpParser\Node\Stmt\Property) {
```

src/Rector/Class_/CommandPropertyToAttributeRector.php
```diff
@@ -0,0 +1,120 @@
...
+            $propertyProperty = $property->props[0];
+            if ($propertyProperty->default instanceof String_) {
+                $defaultName = $propertyProperty->default->value;
+            }
+
+            $this->removeNode($property);
```
samsonasik, line 96: `removeNode()` should only be called when default value is string, so check need to be in if above I guess.

src/Rector/Class_/CommandPropertyToAttributeRector.php
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        $property = $node->getProperty('defaultDescription');
+        if ($property !== null) {
+            $propertyProperty = $property->props[0];
+            if($propertyProperty->default instanceof String_) {
+                $defaultDescription = $propertyProperty->default->value;
+            }
+            $this->removeNode($property);
```
samsonasik, line 105: `removeNode()` should only be called when default value is string, so check need to be in if above I guess.

src/Rector/Class_/CommandPropertyToAttributeRector.php
```diff
@@ -0,0 +1,120 @@
...
+        }
+
```
samsonasik, line 107: `defaultName` and `defaultDescription` need to have initialization variable, so below, we can check if `defaultName` and `defaultDescription` is not null / empty based on initialiazation.

src/Rector/Class_/CommandPropertyToAttributeRector.php
samsonasik:
```suggestion
    if (count($node->props) !== 1) {
        return null;
    }
```

stephanvierkant: I don't get this. Why is this needed?

samsonasik: property can be defined as `private $a, $b, $c;`

TomasVotruba:
I've triggered CI :+1:

TomasVotruba:
I've triggered CI :+1:

src/Rector/Class_/CommandPropertyToAttributeRector.php
stephanvierkant: And why would this be a problem?

samsonasik: removing part of property should not remove whole props if there is other props exists, that may be used, as this rule remove the property itself.

stephanvierkant: Thanks for clarification. I'll fix it!

src/Rector/Class_/CommandPropertyToAttributeRector.php
```diff
@@ -0,0 +1,136 @@
...
+        $property = $node->getProperty('defaultDescription');
+        if ($property instanceof Property) {
+            $propertyProperty = $property->props[0];
+            if ($propertyProperty->default instanceof String_) {
+                $defaultDescription = $propertyProperty->default->value;
+
+                $this->removeNode($property);
+            }
+        }
+
```
samsonasik, line 118: 1 more check:

```suggestion
       if ($defaultDescription === null) {
            return null;
        }
```

stephanvierkant: Description isn't required, is it?

samsonasik: you're correct 👍

samsonasik (approved):

TomasVotruba:
Thank you :+1:  :clap:

PR #165, opened 2022-05-08
Refactor TemplateAnnotationToThisRenderRector to work directly with current stmts

TomasVotruba:


src/NodeFinder/EmptyReturnNodeFinder.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    public function __construct(
+        private readonly BetterNodeFinder $betterNodeFinder
+    ) {
+    }
+
+    public function hasNoOrEmptyReturns(ClassMethod $classMethod): bool
+    {
+        /** @var Return_[] $returns */
+        $returns = $this->betterNodeFinder->findInstanceOf($classMethod, Return_::class);
```
samsonasik, line 22: It should use findInstancesofInFunctionlikeScoped ref https://github.com/rectorphp/rector-src/blob/97a33aa69494836b44ce68e0803e2568233f3e06/src/PhpParser/Node/BetterNodeFinder.php#L447 so it can ignore return in inner function

samsonasik: https://github.com/rectorphp/rector-symfony/pull/166

TomasVotruba: :+1:

PR #166, opened 2022-05-09
[Symfony] Add inner function support on TemplateAnnotationToThisRenderRector

samsonasik:
ref https://github.com/rectorphp/rector-symfony/pull/165#discussion_r867564523


samsonasik:
Fixed 🎉

PR #169, opened 2022-05-13
Add new rule to add annotations from router configuration to symfony controllers

malteschlueter:
See https://github.com/rectorphp/rector-symfony/issues/168

PR #171, opened 2022-05-17
Bugfix: Existing route annotation is duplicated (AddRouteAnnotationRector)

malteschlueter:
We've added in https://github.com/rectorphp/rector-symfony/pull/169 a new rule to add route annoations to controller actions.

Now I could tested it in a "real" application like https://github.com/malteschlueter/symfony-reproducers/tree/feature/rector-add-route-annotation and I found some issues. @TomasVotruba I've installed `rector/rector:dev-main` successfully in the project now :)

---

The first issue is that if a route annotation already exists no matter if the annotation is imported or not. The route annotation will be added again. See in the picture below. Thats why I added the additional tests in the PR which are working correct. Why is the handling in the application different? 🤔 

Edit: Should be fixed with this PR.

grafik

---

The second issue is that the [command](https://github.com/rectorphp/rector/blob/main/vendor/rector/rector-symfony/src/Command/RectorRoutesExportCommand.php#L14) to export the existing routes to a json file is not working without manual changes. The symfony classes has been prefixed and Symfony requires that a command extends from `\Symfony\Component\Console\Command\Command` and it dislikes the prefixed class.

Is there any way to prevent this or do you've any ideas how we improve this?

---

Found a third issue. The arrays will not be added correctly because the code was converted false. It removed the `array_map` completly instead refactor it as a foreach or something else. My idea is to refactor the original code to a `foreach`.

WDYT?
What library is used for converting the code? Maybe we should open an issue for it.

Original-Code: https://github.com/rectorphp/rector-symfony/blob/main/src/Rector/ClassMethod/AddRouteAnnotationRector.php#L75-L82
Converted-Code: https://github.com/rectorphp/rector/blob/main/vendor/rector/rector-symfony/src/Rector/ClassMethod/AddRouteAnnotationRector.php#L68-L77

### DIFF
--- a/src/Rector/ClassMethod/AddRouteAnnotationRector.php
+++ b/src/Rector/ClassMethod/AddRouteAnnotationRector.php
@@ -12,7 +12,6 @@ use Rector\Core\Rector\AbstractRector;
 use Rector\Symfony\DataProvider\RouteMapProvider;
 use Rector\Symfony\PhpDocNode\SymfonyRouteTagValueNodeFactory;
 use Rector\Symfony\ValueObject\SymfonyRouteMetadata;
-use Symfony\Component\Routing\Annotation\Route;
 use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
 
 /**
@@ -59,7 +58,9 @@ class AddRouteAnnotationRector extends AbstractRector
         }
 
         $phpDocInfo = $this->phpDocInfoFactory->createFromNodeOrEmpty($node);
-        $doctrineAnnotationTagValueNode = $phpDocInfo->getByAnnotationClass(Route::class);
+        $doctrineAnnotationTagValueNode = $phpDocInfo->getByAnnotationClass(
+            'Symfony\\Component\\Routing\\Annotation\\Route'
+        );
 
         if ($doctrineAnnotationTagValueNode !== null) {
             return null;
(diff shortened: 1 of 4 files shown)


samsonasik:
Docblock duplication issue possible related with issue 

- https://github.com/rectorphp/rector/issues/7020

I tried to solve it at PR

- https://github.com/rectorphp/rector-src/pull/1862

but still can't found a solution for it. It is a printing issue mostly I think. Feel free to provide a solution if you found it.

malteschlueter:
@samsonasik Thanks for the information 👍

I could fixed the issue with the duplication by changing the following. It happens because it searched for the annotation with a prefixed classname. See https://github.com/rectorphp/rector/blob/main/vendor/rector/rector-symfony/src/Rector/ClassMethod/AddRouteAnnotationRector.php#L61

```diff
@@ -59,7 +58,9 @@ class AddRouteAnnotationRector extends AbstractRector
         }
 
         $phpDocInfo = $this->phpDocInfoFactory->createFromNodeOrEmpty($node);
-        $doctrineAnnotationTagValueNode = $phpDocInfo->getByAnnotationClass(Route::class);
+        $doctrineAnnotationTagValueNode = $phpDocInfo->getByAnnotationClass(
+            'Symfony\\Component\\Routing\\Annotation\\Route'
+        );
```

TomasVotruba:
> I could fixed the issue with the duplication by changing the following.

Awesome :+1:  I'm merging the fix so we have it deployed.

TomasVotruba:
> Now I could tested it in a "real" application like [malteschlueter/symfony-reproducers@feature/rector-add-route-annotation](https://github.com/malteschlueter/symfony-reproducers/tree/feature/rector-add-route-annotation) and I found some issues. @TomasVotruba I've installed rector/rector:dev-main successfully in the project now :)

Thanks for the updat e:tada:

Issue #168, opened 2022-05-12
Feature/Support: Migrate route configuration from yaml to annotation/attribute


malteschlueter:
# Feature Request

I'm actually working on a legacy application with Symfony 3.4 and everything is structured in not necessary bundles. Actually the ~500 routes of the application are configured with YAML in the different bundles. My target is to migrate the route configuration to annotations or attributes. For the beginning I prefer annotations because there are already Rector rules which converts routes between annotations and attributes and legacy applications may not use PHP 8^^

I've already seen that somebody asked for this feature but no one has implemented it or not?

My first step was to get the resolved routes from Symfony because of the bundle structure a route configuration in YAML for a controller/action looks like `FooBundle:Bar:baz` and will be resolved by the router to `\FooBar\FooBundle\Controller\Bar::bazAction` which makes me easily to get the correct controller and especially the action because inside the Controller there are often more than one action implemented. For testing I implemented a Symfony command which iterates through the existing routes from the Symfony router and every information of route like path, name, default, ... are available. Very easy to get the important information.

Now the PHP files must be edited which will be easy with Rector but I don't know if it possible to iterate over the configured routes instead of iterating over directories/files?

If it's not possible maybe it's possible to extend the Symfony command which I used for testing that it can use Rector to edit the PHP files but the isn't my prefered solution because it wouldn't be a contribution to the Rector project but I publish it on GitHub.

WDYT?
I'm very open for any input :) 
 
## Diff 

```diff
-foo_baz:
-    path: /foo/{baz}
-    defaults: { _controller: FooBundle:Bar:baz, _format: json }
-    methods: [ GET, POST ]

+/**
+ * @Route(name="foo_baz", path="/foo/{baz}", methods={"GET","POST"}, defaults={_format="json"})
+ */
public function bazAction(string $baz)
```

TomasVotruba:
Hi, I needed this 2 weeks ago :)

I'd love to see such feature here. I see 2 challanges that needs to be solved:

* provide a path to routing, e.g. via parameter in `$rector->parameters()`
* mapping `FooBundle:Bar:baz` to exact `bazAction()` method

If you solve these 2, we have a great new rule :+1:

malteschlueter:
Does Rector has access to the Symfony container of the application?
I've seen https://github.com/rectorphp/rector-symfony/blob/main/src/ApplicationMetadata/ListenerServiceDefinitionProvider.php but I don't really see through :D

[...]

PR #173, opened 2022-05-21
Refactor AddRouteAnnotationRector to work without command

TomasVotruba:
Ref https://github.com/rectorphp/rector-symfony/issues/168

I've done a little refactoring, mostly about:

* removing dependency on a command, provider service should handle this when the rule is run
* adding routes provider interface to make testing with json easier
* create SymfonyRouteMetadata directly from data provided by Symfony application router
* check if class method is public and non-static, so we can return early and save performance on obviously not controller potential
* tidy up :) 

Could you review [malteschlueter](https://github.com/malteschlueter)?


tests/Rector/ClassMethod/AddRouteAnnotationRector/config/configured_rule.php
```diff
@@ -3,10 +3,19 @@
...
     $rectorConfig->import(__DIR__ . '/../../../../../config/config.php');
+
     $rectorConfig->rule(AddRouteAnnotationRector::class);
-    $rectorConfig->symfonyRoutesJson(__DIR__ . '/route.json');
+
+    // only for tests
+    $services = $rectorConfig->services();
+    $services->set(DummySymfonyRoutesProvider::class);
+
+    // give this service autowiriting preferences
```
samsonasik, line 19: should `autowiriting` be `autowiring` ?

```suggestion
    // give this service autowiring preferences
```

TomasVotruba:
Let's ship it for testing. Feedback is still welcomed :+1:

PR #177, opened 2022-06-01
Add Set Constants that are directed towards the future

dritter:
The current SymfonySetList Constants target the current point in time. With these you can update to the current Symfony version at max.
With this PR I added future directed constants, so that Users don't have to update their rector config if a new Symfony version comes out.

My use-case is:
I have a maintenance bot that does composer updates for me. I would like to see it try Symfony updates as well.


samsonasik:
please run:

```
```

to make CI green.

TomasVotruba:
> The current SymfonySetList Constants target the current point in time. With these you can update to the current Symfony version at max.

This might be handy for current version, but it can be quite surprising when new version is released. E.g. new rules PHP CS Fixer can be really annoying when everything breaks after `composer update`.
Rector should be under control and act only with direct act of user.

PR #180, opened 2022-06-05
[Symfony 2.8] Add FormTypeWithDependencyToOptionsRector

TomasVotruba:
- [Symfony 2.8] Add FormTypeWithDependencyToOptionsRector
- skip no params used

### DIFF
--- a/config/sets/symfony/symfony30.php
+++ b/config/sets/symfony/symfony30.php
@@ -8,6 +8,7 @@ use Rector\Renaming\Rector\MethodCall\RenameMethodRector;
 use Rector\Renaming\Rector\Name\RenameClassRector;
 use Rector\Renaming\ValueObject\MethodCallRename;
 use Rector\Renaming\ValueObject\RenameClassConstFetch;
+use Rector\Symfony\Rector\Class_\FormTypeWithDependencyToOptionsRector;
 use Rector\Symfony\Rector\ClassMethod\FormTypeGetParentRector;
 use Rector\Symfony\Rector\ClassMethod\GetRequestRector;
 use Rector\Symfony\Rector\ClassMethod\RemoveDefaultGetBlockPrefixRector;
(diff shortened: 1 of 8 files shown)


javaDeveloperKid:
Could you descrive the motivation for this Rector into the PR description? I don't understand why one could use this. Why to remove dependencies from form types? Thank you in advance.

bobvandevijver:
@TomasVotruba This rector is crashing for us 

```
 [ERROR] Could not process "src/Form/Type/UserChoiceType.php" file, due to:     
         "System error: "Look at                                                
         "Rector\Symfony\Rector\Class_\FormTypeWithDependencyToOptionsRector::re
         factor()":115"                                                         
         Run Rector with "--debug" option and post the report here:             
         https://github.com/rectorphp/rector/issues/new". On line: 115     
```

This has to do with the form not having the `buildForm` method, which is perfectly fine for a Symfony form.

Our form class:

```php
<?php

namespace App\Form\Type;

use App\Entity\UserProfile;
use Drenso\Shared\Form\Type\Select2EntitySearchType;
use Symfony\Component\Form\AbstractType;
use Symfony\Component\OptionsResolver\Options;
use Symfony\Component\OptionsResolver\OptionsResolver;
use Symfony\Component\Routing\RouterInterface;

class UserChoiceType extends AbstractType
{
  public function __construct(private RouterInterface $router)
  {
  }

  public function configureOptions(OptionsResolver $resolver): void
  {
    $resolver
        ->setDefaults([
            'class'           => UserProfile::class,
            'choice_label'    => 'fullname',
            'placeholder'     => 'user.info.nobody',
        ]);
  }

  public function getParent(): ?string
  {
    return Select2EntitySearchType::class;
  }
}
```

However, a more pressing note: form argument are allowed, and are autowired! So, I'm actually not sure why this rector is being added to the Symfony30 set, which is included in the set we use (symfony60)...

PR #189, opened 2022-06-20
Fix FormTypeWithDependencyToOptionsRector configuration

bobvandevijver:
As asked in #180, but I have no clue on how to make a correct failing test fixture for this :(

### DIFF
--- a/config/sets/symfony/symfony30.php
+++ b/config/sets/symfony/symfony30.php
@@ -8,7 +8,6 @@ use Rector\Renaming\Rector\MethodCall\RenameMethodRector;
 use Rector\Renaming\Rector\Name\RenameClassRector;
 use Rector\Renaming\ValueObject\MethodCallRename;
 use Rector\Renaming\ValueObject\RenameClassConstFetch;
-use Rector\Symfony\Rector\Class_\FormTypeWithDependencyToOptionsRector;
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Thank you :+1: 

Each rule has their test with same-named directory.
https://github.com/rectorphp/rector-symfony/tree/main/tests/Rector/Class_/FormTypeWithDependencyToOptionsRector

Test fixture is simply a file in `/Fixture` directory.

bobvandevijver:
Yeah, that I found, but I'm not sure how to create a failing test for the incorrect set inclusion? Or are you looking for the test for when the `buildForm` is missing?

TomasVotruba:
> I tried something, but I failed to run the test locally due to

Try running `composer install`. Should apply the patch to make this go away :+1: 

> Well, the test is failing as it should slightly_smiling_face

Look good, thanks :slightly_smiling_face: I'll be working on Symfony Twig rules today, so I'll skip make sure this case is fixed.

bobvandevijver:
> Try running `composer install`. Should apply the patch to make this go away 👍

I did indeed, my bad, by default I run the install without scripts/plugins to speed up the installation 😃 

> Look good, thanks 🙂 I'll be working on Symfony Twig rules today, so I'll skip make sure this case is fixed.

Alright 👍🏻

PR #193, opened 2022-06-23
Bugfix: Allow nullable for ControllerReference (AddRouteAnnotationRector)

malteschlueter:
Today I've found a new edge case which occurs a error in the `AddRouteAnnotationRector` rule.

If you define a route with a path but without a controller (it's allowed in Symfony) the `Rector\Symfony\ValueObject\SymfonyRouteMetadata::getControllerReference()` method throws the error `Return value must be of type string, null returned`.

### DIFF
--- a/src/NodeAnalyzer/Annotations/MethodCallAnnotationAssertResolver.php
+++ b/src/NodeAnalyzer/Annotations/MethodCallAnnotationAssertResolver.php
@@ -34,9 +34,9 @@ final class MethodCallAnnotationAssertResolver
 
         // based on https://github.com/symfony/symfony/blob/7d4b42cbeef195e0a01272b9c5f464f0afe52542/src/Symfony/Component/Validator/Mapping/GetterMetadata.php#L45-L47
         $possibleMethodNames = [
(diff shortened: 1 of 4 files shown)


samsonasik:
you can run `vendor/bin/rector && composer fix-cs` to make CI green.

samsonasik:
@TomasVotruba @malteschlueter that's due to `ucfirst` added to PR 

- https://github.com/rectorphp/rector-src/pull/2543

by @hungtrinh.

`$firstArgValue` can be non-string, so marking as `@var string` can be used for this:

```php
/* @var string $propertyName */
$propertyName = $this->valueResolver->getValue($firstArgValue);
```

TomasVotruba:
@samsonasik Ah, I see. I thought it's a strict type of `string`, but it's not.
https://github.com/rectorphp/rector-symfony/pull/193/files#diff-107058072837039f859dd7a903d7d04892e13ba80e60cdd412070900757cfeecR33

Thanks for clarification, it's correct change then :+1:

TomasVotruba:
@malteschlueter Thank you :+1:

PR #198, opened 2022-06-28
Add missing "options" in routes (AddRouteAnnotationRector)

malteschlueter:
When I used the `AddRouteAnnotationRector` rule, I noticed that the options of a route are also important and are missing in the rule.

The Route class sets by default a `compiler_class` value as you see here https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/Routing/Route.php#L253. Because of that I ignore this explicit but allow if somebody changed it because the RouteCompiler is based on a interface https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/Routing/RouteCompiler.php#L20.


TomasVotruba:
Thank you :+1:

PR #199, opened 2022-06-28
Fix documentation on AuthorizationCheckerIsGrantedExtractorRector

romainjanvier:
Improve doc formatting and fix duplicated ROLE_USER instead of ROLE_ADMIN


samsonasik:
The part that need update is ini the `AuthorizationCheckerIsGrantedExtractorRector` itself as doc is generated with command `composer docs`

romainjanvier:
Thanks for tips, I amended my commit

samsonasik (approved):

TomasVotruba:
Thank you :+1:

PR #207, opened 2022-07-21
[Symfony60] Fix added return type for Router::getRouteCollection

kick-the-bucket:
Fixes rectorphp/rector#7308


TomasVotruba:
Looks very good, thank you for the test verification :clap:

PR #212, opened 2022-08-05
Add a failing fixtures for RemoveDefaultGetBlockPrefixRector

lyrixx:
ref: https://github.com/rectorphp/rector/issues/7347

### DIFF
--- a/tests/Rector/ClassMethod/RemoveDefaultGetBlockPrefixRector/Fixture/skip_with_get_block_prefix.php.inc
+++ b/tests/Rector/ClassMethod/RemoveDefaultGetBlockPrefixRector/Fixture/skip_with_get_block_prefix.php.inc
@@ -1,0 +1,23 @@
+<?php
+
+namespace Rector\Symfony\Tests\Rector\ClassMethod\RemoveDefaultGetBlockPrefixRector\Fixture;
+
+use Symfony\Component\Form\AbstractType;
+
+class IconType extends ChoiceType
+{
+    public function getBlockPrefix()
+    {
+        return 'icon';
+    }
+}
+
+class ChoiceType extends AbstractType
+{
+    public function getBlockPrefix()
+    {
+        return 'something';
+    }
+}
+
+?>


tests/Rector/ClassMethod/RemoveDefaultGetBlockPrefixRector/Fixture/extends.php.inc
```diff
@@ -0,0 +1,47 @@
...
+    {
+        return 'icon';
+    }
+}
+
+class ChoiceType extends AbstractType
+{
+    public function getBlockPrefix()
+    {
+        return 'choice';
```
lyrixx, line 19: In SF 3.4, the parent (ChoiceType) **has** this method. [ref](https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/Form/Extension/Core/Type/ChoiceType.php#L357-L360)

So If I remove the method in **my** type, I will end up with "choice" and not the default implement in https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/Form/AbstractType.php#L53-L56

tests/Rector/ClassMethod/RemoveDefaultGetBlockPrefixRector/Fixture/extends.php.inc
```diff
@@ -0,0 +1,47 @@
...
+    }
+}
+
+?>
```
samsonasik, line 47: you can remove second part if before and after is equal, also, please rename fixture file with `skip_` prefix, eg: `skip_with_get_block_prefix.php.inc`

lyrixx: This fixture is wrong. It's just an example requested by https://github.com/rectorphp/rector/issues/7347

samsonasik: What is the right one?

lyrixx: It's complicated! see this [comment](https://github.com/rectorphp/rector-symfony/pull/212#discussion_r938984638)


But I have an idea!

lyrixx: Pushed, the test is now failing as expected, and the fixtures is OK, no need to update it.

TomasVotruba:
I've cherry-picked your test fixture here https://github.com/rectorphp/rector-symfony/pull/219

PR #213, opened 2022-08-06
Add ServiceSetStringNameToClassNameRector for Symfony 2.8 fluent configs

TomasVotruba:


samsonasik:
PHPStan notice is due to missing stubs for classes to detect string is a class-string, ref 

- https://github.com/rectorphp/rector-src/pull/2726

It can temporary registered to PHPStan ignore config:

```neon
        # this needs stubs for classes to detect string is a class-string
        # @see https://github.com/rectorphp/rector-src/pull/2726
        -
            message: '#Parameter \#\d \$.+ of class .+ constructor expects class\-string, string given#'
            path: config
```

TomasVotruba:
This invalid `class-string` actually should be reverted: https://github.com/rectorphp/rector-src/pull/2739
I didn't realize the PHPStan sees non-existing classes as bare `string`, so having `class-string` does not make sense.

samsonasik:
Close-Reopen to restart build as rectify hang on rector-* packages after auto commit

TomasVotruba:
Not neccesasy, this is WIP. I'll handle it

TomasVotruba:
Inspiration from https://github.com/symplify/symplify/pull/4356

PR #217, opened 2022-08-08
Move SymfonyLevelSetList constants to SymfonySetList

### LINKED ISSUE #216: Merge SymfonySetList and SymfonyLevelSetList
I personally struggle all the time getting the correct import for the `UP_TO_SYMFONY_61` rule sets.
Would find it better if the `SetList` and `LevelSetList` are merged into a single imported file so I find them at the same place. As sometimes I switching from `UP_TO_*` to single version and vise-versa when doing step by step updates, but have then always import the one or the other file make it a better developer experience from my point of view.

alexander-schranz:
Fixes #216

### DIFF
--- a/src/Set/SymfonyLevelSetList.php
+++ b/src/Set/SymfonyLevelSetList.php
@@ -6,6 +6,9 @@ namespace Rector\Symfony\Set;
 
 use Rector\Set\Contract\SetListInterface;
 
+/**
+ * @deprecated Use the constants of SymfonySetList instead.
+ */
 final class SymfonyLevelSetList implements SetListInterface
 {
     /**
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ return static function (RectorConfig $rectorConfig): void {
     $rectorConfig->symfonyContainerXml(__DIR__ . '/var/cache/dev/App_KernelDevDebugContainer.xml');
 
     $rectorConfig->sets([
-        SymfonySetList::SYMFONY_52,
+        SymfonySetList::SYMFONY_60, // to run also previous symfony version ruleSets use: SymfonySetList::UP_TO_SYMFONY_60
         SymfonySetList::SYMFONY_CODE_QUALITY,
         SymfonySetList::SYMFONY_CONSTRUCTOR_INJECTION,
     ]);
--- a/config/sets/symfony/level/up-to-symfony-26.php
+++ b/config/sets/symfony/level/up-to-symfony-26.php
@@ -3,9 +3,8 @@
 declare(strict_types=1);
 
 use Rector\Config\RectorConfig;
-use Rector\Symfony\Set\SymfonyLevelSetList;
 use Rector\Symfony\Set\SymfonySetList;
 
 return static function (RectorConfig $rectorConfig): void {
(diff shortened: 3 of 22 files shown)


src/Set/SymfonyLevelSetList.php
```diff
@@ -6,6 +6,9 @@
 
 use Rector\Set\Contract\SetListInterface;
 
+/**
+ * @deprecated Use the constants of SymfonySetList instead.
+ */
```
alexander-schranz, line 11: I currently deprecated it not sure when you planing the next release with breaking changes.

alexander-schranz:
Think the CI errors are unrelated to the PR.

samsonasik:
It seems it make cimplicated compared with original, per-naming, the purpose is clear, with this, it now become ambiguous

alexander-schranz:
@samsonasik not understanding if you think its a bad idea or not 🙈

samsonasik:
I think it is bad idea. The existing value objects: `LevelSetList` vs `SetList` already clear. If there is a bug on range, we know when to search, combine it will make it harder.

alexander-schranz:
Okay, must say I have another view on the topic and see that the `Level` classes add here no value and a `Single` class would make it easier to access all Sets.

samsonasik:
imo, if it needs improvement, should be on documentation side (readme or docs) for how differentiate between LevelSetList vs SetList.

alexander-schranz:
Lets see what @TomasVotruba means. I personally see no value in `Level` classes as the constants names itself are self documenting and merging them into a single class would avoid confusion and so avoid to document why LevelSets are for and what other Sets are for.

samsonasik:
There will be many things to change because of this in various rector-* packages: symfony, cakephp, laravel, phpunit, rector-src up level setlist, downgrade-rector-php, for something that unnecessary and increase complexity and make maintenance harder, deprecating things, and need to remove in the future, change everything in every packages, make announcement, etc...

The purpose of decoupling are: easy to find, easy to fix, more maintainable.

alexander-schranz:
> There will be many things to change because of this in various rector-* packages: symfony, cakephp, laravel, phpunit, rector-src up level setlist, downgrade-rector-php, for something that unnecessary and increase complexity and make maintenance harder, deprecating things, and need to remove in the future, change everything in every packages, make announcement, etc...

More an argument better do it know then later :) I already have the other repos on my list just waiting for getting feedback if we want to deprecate it currently or remove the class if we can do it know. If we go with deprecate it and removing it later, I can also create the remove pull requests if there will be a 0.14 on the repositories.

TomasVotruba:
My apologies for late interaction here :) I got busy with Rector 0.14 release.

On second thought, the added value of single class is less classes, but ambiguous meaning.
From the all comments above, the @samsonasik resonates the most with purpose of these set classes:

> I think it is bad idea. The existing value objects: LevelSetList vs SetList already clear. If there is a bug on range, we know when to search, combine it will make it harder.

When I look for potential single set to include all the lower ones to the highest, I type "LevelSet*"  and see what is available. Not all vendors have this option, so looking united naming makes it simple.

Saying that, let's keep the sets as they're are :+1: 

@alexander-schranz Thanks for understanding and even proposal.

PR #218, opened 2022-08-12
[Symfony] Handle single action with __construct on InvokableControllerRector

samsonasik:
Given the following code:

```php
use Symfony\Component\HttpFoundation\Response;
use Symfony\Component\Serializer\SerializerInterface;

final class SingleActionWithConstruct extends AbstractController
{
    public function __construct(
        private SerializerInterface $serializer,
    ) {
    }

    public function create(): Response
    {
        return $this->json([], Response::HTTP_CREATED);
    }
}
```

It currently produce error:

```bash

There was 1 error:

1) Rector\Symfony\Tests\Rector\Class_\InvokableControllerRector\InvokableControllerRectorTest::test with data set #1 (Symplify\SmartFileSystem\SmartFileInfo Object (...))
Undefined array key 0
```

Fixes https://github.com/rectorphp/rector/issues/7365

### DIFF
--- a/src/Rector/Class_/InvokableControllerRector.php
+++ b/src/Rector/Class_/InvokableControllerRector.php
@@ -112,7 +112,10 @@ CODE_SAMPLE
 
         // 1. single action method → only rename
         if (count($actionClassMethods) === 1) {
-            return $this->refactorSingleAction($actionClassMethods[0], $node);
+            return $this->refactorSingleAction(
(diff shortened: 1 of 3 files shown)


samsonasik:
Fixed 🎉 /cc @Myks92

samsonasik:

TomasVotruba:
Thank you :+1:

PR #223, opened 2022-08-19
Handle static calls in WebTestCaseAssertResponseCodeRector

Khartir:
Both PHPUnit's and Symfony's `assert*` methods are `static`, so this adds the handling of static calls.

### DIFF
--- a/src/Rector/MethodCall/WebTestCaseAssertResponseCodeRector.php
+++ b/src/Rector/MethodCall/WebTestCaseAssertResponseCodeRector.php
@@ -7,6 +7,7 @@ namespace Rector\Symfony\Rector\MethodCall;
 use PhpParser\Node;
 use PhpParser\Node\Expr;
 use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\StaticCall;
 use PHPStan\Type\ObjectType;
 use Rector\Core\Rector\AbstractRector;
 use Rector\PHPUnit\NodeAnalyzer\TestsNodeAnalyzer;
@@ -68,11 +69,11 @@ CODE_SAMPLE
      */
     public function getNodeTypes(): array
     {
-        return [MethodCall::class];
+        return [MethodCall::class, StaticCall::class];
     }
 
     /**
-     * @param MethodCall $node
+     * @param MethodCall|StaticCall $node
      */
     public function refactor(Node $node): ?Node
     {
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Thank you :+1: 

I've triggered the CI to run tests.

src/Rector/MethodCall/WebTestCaseAssertResponseCodeRector.php
```diff
@@ -148,10 +152,14 @@ private function processAssertResponseStatusCodeSame(MethodCall $methodCall): ?M
...
 
+        if ($methodCall instanceof StaticCall) {
+            return $this->nodeFactory->createStaticCall('self', 'assertResponseStatusCodeSame', [$methodCall->args[0]]);
+        }
+
         return $this->nodeFactory->createLocalMethodCall('assertResponseStatusCodeSame', [$methodCall->args[0]]);
     }
 
-    private function processAssertResponseRedirects(MethodCall $methodCall): ?MethodCall
+    private function processAssertResponseRedirects(Expr\CallLike $methodCall): ?Expr\CallLike
```
samsonasik, line 162: using union for exact types should can be applied:

```suggestion
    private function processAssertResponseRedirects(MethodCall|StaticCall $methodCall): MethodCall|StaticCall|null
```

src/Rector/MethodCall/WebTestCaseAssertResponseCodeRector.php
```diff
@@ -111,7 +112,10 @@ public function isGetLocationMethodCall(Expr $expr): bool
         return $this->valueResolver->isValue($firstArg->value, 'Location');
     }
 
-    private function processAssertResponseStatusCodeSame(MethodCall $methodCall): ?MethodCall
+    /**
+     * @param StaticCall|MethodCall $methodCall
+     */
+    private function processAssertResponseStatusCodeSame(Expr\CallLike $methodCall): ?Expr\CallLike
```
samsonasik, line 118: using union for exact types should can be applied:

```suggestion
    private function processAssertResponseStatusCodeSame(StaticCall|MethodCall $methodCall): MethodCall|StaticCall|null
```

src/Rector/MethodCall/WebTestCaseAssertResponseCodeRector.php
Khartir: For some reason I thought I couldn't use this for backwards compatibility, but I just never checked.

samsonasik:
Please run fix cs and rector to make CI green:

```
composer fix-cs && vendor/bin/rector
```

samsonasik:
Please run once more:

```
composer fix-cs
```

TomasVotruba:
This is good enough. The Rector and ECS here are only for style, we can handle it in next PR.
The functionality passes well :heavy_check_mark: 

Thank you

Khartir:
> Please run once more:
> 
> ```
> composer fix-cs
> ```
> 
> my command was flipped as fix-cs should be later.

To be honest, I'm not sure what happened here. `vendor/bin/rector` doesn't do anything locally, so I must have accidentally put a new line back in or something?

samsonasik:
Probably, its green now, so it is ok 👍

samsonasik (approved):

TomasVotruba:
Awesome :) thank you both :+1:

PR #224, opened 2022-08-19
Add rector to replace KernelTestCase::$container with KernelTestCase::getContainer() in tests

Khartir:
Symfony deprecated `KernelTestCase::$container` back in `5.3` (https://github.com/symfony/symfony/blob/5.3/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php#L39) and removed it in `6.0`.

The recommended alternative ist to call `KernelTestCase::getContainer()` instead.


TomasVotruba:
Thank you for the contribution :+1:  LGTM

Could you register the rule in specific Symfony set? So the change is included during Symfony set run.

Khartir:
Do you mean moving it from the Symfony 6 set to the Symfony 5.3 set? Or which set should I add this to?

TomasVotruba:
Oh, my bad. I though it's a new rule :)

TomasVotruba:
Thank you :clap:

PR #235, opened 2022-09-01
[Dep] Update tests to use string path instead of SmartFileInfo

samsonasik:
ref https://github.com/rectorphp/rector-src/pull/2889

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -20,7 +20,7 @@ return static function (RectorConfig $rectorConfig): void {
         '*/Source/*',
         '*/Source*/*',
         '*/tests/*/Fixture*/Expected/*',
-        StringClassNameToClassConstantRector::class => [__DIR__ . '/config'],
+        StringClassNameToClassConstantRector::class,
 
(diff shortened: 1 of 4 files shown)


TomasVotruba:
I'm going off soon, feel free to merge these 👍

src/Rector/ClassConstFetch/ConsoleExceptionToErrorEventConstantRector.php
```diff
@@ -57,10 +59,13 @@ public function refactor(Node $node): ?Node
     {
         if ($node instanceof ClassConstFetch && (
             $this->isObjectType($node->class, $this->consoleEventsObjectType) &&
-            $this->isName($node->name, 'EXCEPTION')
+            $this->isName($node->name, EXCEPTION::class)
```
samsonasik, line 62: is seems new bug on `StringClassNameToClassConstantRector`, need to skip it for now.

samsonasik: https://github.com/rectorphp/rector-symfony/pull/236

PR #236, opened 2022-09-02
Exclude Error and Exception class name on StringClassNameToClassConstantRector

samsonasik:
ref https://github.com/rectorphp/rector-symfony/pull/235#pullrequestreview-1094097162

I think the skip config should actually merge instead of replace existing `classesToSkip` config.


TomasVotruba:
👍

PR #238, opened 2022-09-11
Add failing test for Command::execute return int rector with try catch

acrobat:
Add failing test for ConsoleExecuteReturnIntRector when execute method has a try catch setup and contains return's in the try and catch part so no extra return should be necessary.

See example: https://getrector.org/demo/7d04f7ed-a69a-4fd1-8389-c1444e3ba15d


samsonasik:
@acrobat there is RemoveUnreachableStatementRector that cover that, you can combine the rule with that, see https://getrector.org/demo/9be494b3-c105-440f-9035-8065f684b470

TomasVotruba:
@samsonasik I think this case should be skipped in this rule too, as the return value is always integer.

samsonasik:
@TomasVotruba Ok 👍 , @acrobat I cherry-picked your commit at PR https://github.com/rectorphp/rector-symfony/pull/239

acrobat:
Thanks for the quick fix @samsonasik!

Issue #229, opened 2022-08-22
setDescription not changed for Commands MakeCommandLazyRector|CommandPropertyToAttributeRector


johanadivare:
When i have a `setDescription` in my `configure` function it is not moved to static property for https://github.com/rectorphp/rector-symfony/blob/main/docs/rector_rules_overview.md#makecommandlazyrector and not moved when i enable https://github.com/rectorphp/rector-symfony/blob/main/docs/rector_rules_overview.md#commandpropertytoattributerector

TomasVotruba:
Thank you for your report!

We'll need an isolated failing demo link from: http://getrector.org/demo, that way we can reproduce the bug.

johanadivare:
it would look like this:
https://getrector.org/demo/8bd44b67-1f32-4f36-b00a-d1b96a092506

But its says no changes while it is changing it for me like:
```php
<?php

declare(strict_types=1);

namespace App\Command;

use Symfony\Component\Console\Command;
use Symfony\Component\Console\Input\InputInterface;
use Symfony\Component\Console\Output\OutputInterface;

final class TestCommand extends Command
{
    protected static $defaultName = 'cron:test';
    /**
     * {@inheritdoc}
     */
    protected function configure(): void
    {
        $this->setDescription('testing');
    }

    /**
     * {@inheritdoc}
     */
    protected function execute(InputInterface $input, OutputInterface $output): int
    {
        return 0;
    }
}
```

TomasVotruba:
Thanks :+1: I see many sets there in `rector.php`. It should be only single rule that handles the changes. Could you update it?

TomasVotruba:
I see it now, thanks :+1: 

Will look into it

TomasVotruba:
What Symfony version do you use BTW?

Seem the issue is in wrong Command class name:

```diff
-use Symfony\Component\Console\Command;
-+use Symfony\Component\Console\Command\Command;
```

johanadivare:
Ah good catch! 

But now my problem with https://getrector.org/demo/f261ad24-5764-4143-a6f4-ad78bc1fa34b is that there is still a `setDescription` i would expect a protected static property for it. Same case with the other config https://getrector.org/demo/eb3a6df7-061b-4001-9fa4-b1b25713110d.

Also when i use https://getrector.org/demo/5b8e8d18-26d2-4a72-bef4-629109c3a5c8 there is no change while i also expect them to change to attribute right?

[...]

PR #245, opened 2022-09-14
[Symfony53] Add CommandDescriptionToPropertyRector

JohJohan:
See https://github.com/rectorphp/rector-symfony/issues/229

### DIFF
--- a/config/sets/symfony/symfony53.php
+++ b/config/sets/symfony/symfony53.php
@@ -10,6 +10,7 @@ use Rector\Renaming\Rector\MethodCall\RenameMethodRector;
 use Rector\Renaming\Rector\Name\RenameClassRector;
 use Rector\Renaming\ValueObject\MethodCallRename;
 use Rector\Renaming\ValueObject\RenameClassConstFetch;
+use Rector\Symfony\Rector\Class_\CommandDescriptionToPropertyRector;
 use Rector\Symfony\Rector\StaticPropertyFetch\KernelTestCaseContainerPropertyDeprecationRector;
 use Rector\Symfony\Set\SymfonySetList;
 use Rector\TypeDeclaration\Rector\ClassMethod\AddReturnTypeDeclarationRector;
@@ -99,4 +100,5 @@ return static function (RectorConfig $rectorConfig): void {
         ]);
 
     $rectorConfig->rule(KernelTestCaseContainerPropertyDeprecationRector::class);
+    $rectorConfig->rule(CommandDescriptionToPropertyRector::class);
 };
--- a/src/Rector/Class_/CommandDescriptionToPropertyRector.php
+++ b/src/Rector/Class_/CommandDescriptionToPropertyRector.php
@@ -1,0 +1,172 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Symfony\Rector\Class_;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Stmt\Class_;
+use PhpParser\Node\Stmt\Property;
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\StringType;
+use Rector\Core\NodeAnalyzer\ExprAnalyzer;
+use Rector\Core\Rector\AbstractRector;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+/**
+ * @see https://symfony.com/blog/new-in-symfony-5-3-lazy-command-description
+ *
+ * @see \Rector\Symfony\Tests\Rector\Class_\CommandDescriptionToPropertyRector\CommandDescriptionToPropertyRectorTest
+ */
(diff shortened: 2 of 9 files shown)


TomasVotruba:
Thanks :+1: 

What Symfony version support the name and the description?

JohJohan:
I think from 5.3 they support this: https://symfony.com/doc/5.3/console.html (but that is not maintained) so i would say 5.4/6.0 and 6.1

The blog post you probably found: https://symfony.com/blog/new-in-symfony-5-3-lazy-command-description

What do you mean by 

> thus 2 different configs and rules.

Should i use different configs to test this behavior? And is there some docs about that?

TomasVotruba:
That is for the description, yes :+1: 

The `$defaultName` is available since Symfony 3.4: https://github.com/symfony/symfony/pull/23887

JohJohan:
Ah like that i can do that but takes a bit more time

config/sets/symfony/symfony53.php
```diff
@@ -99,4 +100,5 @@
         ]);
 
     $rectorConfig->rule(KernelTestCaseContainerPropertyDeprecationRector::class);
+    $rectorConfig->rule(CommandDescriptionToPropertyRector::class);
```
JohJohan, line 103: @TomasVotruba i think this is desired because the feature was added after 5.3?

src/Rector/Class_/CommandDescriptionToPropertyRector.php
```diff
@@ -0,0 +1,146 @@
...
+}
+CODE_SAMPLE
+                ,
+                <<<'CODE_SAMPLE'
+use Symfony\Component\Console\Command\Command
+
+final class SunshineCommand extends Command
+{
+    protected static $defaultName = 'sunshine';
+    protected static $defaultDescription = 'sunshine description';
```
JohJohan, line 49: Is there a way to force this order? Right now its adding the defaultDescription as first property.
So first `$defaultName` and then `$defaultDescription`

src/Rector/Class_/CommandDescriptionToPropertyRector.php
```diff
@@ -0,0 +1,146 @@
...
+        if ($defaultNameProperty instanceof Property) {
+            return null;
+        }
+
+        $commandDescription = $this->resolveCommandDescription($node);
+        if (! $commandDescription instanceof Node) {
+            return null;
+        }
+
+        if (! $commandDescription instanceof String_ && ! $commandDescription instanceof ClassConstFetch) {
```
samsonasik, line 86: ClassConstFetch can be dynamic value, you can verify with ExprAnalyzer, see https://github.com/rectorphp/rector-src/blob/73aa7e3d68d74fc0b9b26fc543bf1207cfc8d0dd/src/NodeAnalyzer/ExprAnalyzer.php#L69

src/Rector/Class_/CommandDescriptionToPropertyRector.php
TomasVotruba: You'd have to look for existing properties on the `Class_` and if there's a property of "defaultName", add a new property after it.

src/Rector/Class_/CommandDescriptionToPropertyRector.php
JohJohan: I dont realy understand what i have to do. Are you saying i have to check that the `$commandDescription` is `isDynamicExpr` otherwise return null? like:
```suggestion
        if (! $commandDescription instanceof String_ && ! $commandDescription instanceof ClassConstFetch && ! $this->exprAnalyzer->isDynamicExpr($commandDescription)) {
```


I used this code as example btw https://github.com/rectorphp/rector-symfony/blob/main/src/Rector/Class_/MakeCommandLazyRector.php#L97 where the same check is done.

samsonasik: You should can use return null early when it is dynamic:

```php

if (! $commandDescription instanceof Expr) {
    return null;
}

if ($this->exprAnalyzer->isDynamicExpr($commandDescription)) {
    return null;
}
```

You need to inject `Rector\Core\NodeAnalyzer\ExprAnalyzer` to constructor first for that:

```php
public function __construct(private readonly \Rector\Core\NodeAnalyzer\ExprAnalyzer $exprAnalyzer)
{
}
```

JohJohan: Check i added this check and also removed check on Node as that was doing nothing anymore

samsonasik:
You can run:

```
```

To make CI green.

JohJohan:
Thanks @samsonasik i did and squashed my commits into 1 commit.

src/Rector/Class_/CommandDescriptionToPropertyRector.php
samsonasik: String_ and ClassConstFetch check no longer needed, can be removed.

samsonasik:
String_ and ClassConstFetch check no longer needed, can be removed as already checked by ExprAnalyzer.

Also, please update PR Title to something like:

```
[Symfony53] Add CommandDescriptionToPropertyRector

samsonasik (approved):

TomasVotruba:
Thank you :clap:

Issue #255, opened 2022-10-03
Use Symfony Command constants for return of execute


johanadivare:
With https://github.com/rectorphp/rector-symfony/blob/main/docs/rector_rules_overview.md#consoleexecutereturnintrector we got explicit return type int but it is even nicer to use Symfony Command constants they also use it as example in there documentation see https://symfony.com/doc/current/console.html#creating-a-command

Example:
```diff
 class SomeCommand extends Command
 {
    public function execute(InputInterface $input, OutputInterface $output): int
     {
-         return 0;
+        return Command::SUCCESS;
     }
 }
```

If you say its a nice addition a new rule should be added as its not part off `ConsoleExecuteReturnIntRector` right?

TomasVotruba:
This would be great addition :+1:

johanadivare:
Okay i will work on this

johanadivare:
Fixed with https://github.com/rectorphp/rector-symfony/pull/256

PR #256, opened 2022-10-04
[CommandConstantReturnCodeRector] Add CommandConstantReturnCodeRector (#255)

JohJohan:
See #255


src/Rector/Return_/CommandConstantReturnCodeRector.php
```diff
@@ -82,19 +82,31 @@ public function refactor(Node $node): ?Node
...
+        if (! $this->nodeNameResolver->isName($node, 'execute')) {
             return null;
         }
 
-        if (! $node->expr instanceof LNumber) {
-            return null;
-        }
-
-        $node->expr = $this->convertNumberToConstant($node->expr);
+        $this->findReturnStatement($node->stmts);
```
samsonasik, line 89: The `Return_` to be found must be scoped, not under Closure or inner function, so you need to use : 

```
$this->betterNodeFinder->findInstancesOfInFunctionLikeScoped($node, [Return_::class]);
```

src/Rector/Return_/CommandConstantReturnCodeRector.php
```diff
@@ -65,7 +65,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
      */
     public function getNodeTypes(): array
     {
-        return [Return_::class];
+        return [ClassMethod::class];
```
samsonasik, line 68: namespace must be changed to under `namespace Rector\Symfony\Rector\ClassMethod`

src/Rector/Return_/CommandConstantReturnCodeRector.php
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * @return array<class-string<Node>>
+     */
+    public function getNodeTypes(): array
+    {
+        return [ClassMethod::class];
+    }
+
+    /**
+     * @param Return_ $node
```
samsonasik, line 72:
```
@param ClassMethod $node
```

src/Rector/Return_/CommandConstantReturnCodeRector.php
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        if (! $classReflection->isSubclassOf('Symfony\Component\Console\Command\Command')) {
+            return null;
+        }
+
+        if (! $classReflection->hasMethod('execute')) {
+            return null;
+        }
+
```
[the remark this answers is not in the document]
johanadivare: Your right i added `skip_already_using_constant_and_other_return.php.inc` to check it only replaces correct return types.

src/Rector/Return_/CommandConstantReturnCodeRector.php
johanadivare: Will have a look again to see if i can fix it with `findInstancesOfInFunctionLikeScoped`

johanadivare: Only `replace_with_constant_nested_function.php.inc` doesn't work as the `expr` on line 95 is a `MethodCall` to `executeCommandStuff`

src/Rector/Return_/CommandConstantReturnCodeRector.php
johanadivare: Fixed

src/Rector/Return_/CommandConstantReturnCodeRector.php
johanadivare: @samsonasik maybe you can give a hint?

samsonasik: @johanadivare that's because you already have this skip:

```php
        if (! $this->nodeNameResolver->isName($node, 'execute')) {
            return null;
        }
```

that make it skipped, as the method you want to change the return value is different method.

You may need separate Rector rule for that, that will be easier to verify the method is called in `Class_`->getMethod('execute')`

tests/Rector/ClassMethod/CommandConstantReturnCodeRector/Fixture/skip_other_function_call_return_int.php.inc
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Rector\Symfony\Tests\Rector\ClassMethod\CommandConstantReturnCodeRector\Fixture;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class SkipAlreadyUsingConstantNestedFunction extends Command
```
samsonasik, line 9: this fixture can be removed, out of rule context. 

New Rector rule should be used to verify if private method called in `execute()`

src/Rector/Return_/CommandConstantReturnCodeRector.php
JohJohan: Okay fair point, i will remove the part checking for nested functions

tests/Rector/ClassMethod/CommandConstantReturnCodeRector/Fixture/skip_other_function_call_return_int.php.inc
JohJohan: ++

samsonasik (approved):

TomasVotruba:
Thank you :clap:

Issue #254, opened 2022-10-03
[CommandPropertyToAttributeRector] is not setting all attributes


johanadivare:
# Bug Report
CommandPropertyToAttributeRector is not setting `aliases` and `hidden` in `AsCommand` attribute

| Subject        | Details              |
| :------------- | :--------------------|
| Rector version | last dev-main      |
| Installed as   | composer dependency  |

## Minimal PHP Code Causing Issue

See https://getrector.org/demo/69d11564-5a50-458b-b185-de0c343b416b

```php
<?php

class SunshineCommand extends \Symfony\Component\Console\Command\Command
{
    /** @var string|null */
    public static $defaultName = 'sunshine';

    /** @var string|null */
    public static $defaultDescription = 'Let the sun shine!';
    

    /**
     * {@inheritdoc}
     */
    protected function configure(): void
    {
        $this
            ->setAliases(['sun', 'shine'])
        ;
    }
}
```

### Responsible rules

* `CommandPropertyToAttributeRector`

## Expected Behavior
```php
<?php

#[\Symfony\Component\Console\Attribute\AsCommand('sunshine', 'Let the sun shine!', ['sun', 'shine'], false)
class SunshineCommand extends \Symfony\Component\Console\Command\Command
{
    /**
     * {@inheritdoc}
     */
    protected function configure(): void
    {
    }
}
```

johanadivare:
Maybe a new rule is more appropriate as the `aliases` and `hidden` are not set with properties but set with configure but they are properties in `\Symfony\Component\Console\Command\Command`

johanadivare:
@JohJohan have a look

stof:
Also, the `MakeCommandLazyRector` is not migrating the description to `$defaultDescription` (probably because it is a lot newer than the `$defaultName`) and so it misses it in most cases too.

JohJohan:
@stof this is done with https://github.com/rectorphp/rector-symfony/pull/257 as discussed here https://github.com/rectorphp/rector-symfony/issues/229

stof:
the rector present in that PR is not documented, which is why I missed it

JohJohan:
@stof good point, i created a pull request to add it to the docs: https://github.com/rectorphp/rector-symfony/pull/260

johanadivare:
Wip: https://github.com/rectorphp/rector-symfony/pull/259

johanadivare:
Fixed with https://github.com/rectorphp/rector-symfony/pull/259

PR #259, opened 2022-10-24
[CommandPropertyToAttribute] #254

JohJohan:


johanadivare:
The test are failing on `Rector\Core\Exception\ShouldNotHappenException: Chain method calls cannot be removed this way. It would remove the whole tree of calls. Remove them manually by creating new parent node with no following method.` i dont know how to remove the node for hidden setter

JohJohan:
@samsonasik do you know why https://github.com/rectorphp/rector-symfony/pull/259#issuecomment-1289051683?

samsonasik:
@JohJohan you were trying to remove partial method call from fluent call. You need to create new parent node or replace root node with new Node instead.

JohJohan:
> @JohJohan you were trying to remove partial method call from fluent call. You need to create new parent node or replace root node with new Node instead.

Okay, i looked into it but i'm affraid i'm not able to fix it. I tried looking at our docs or other code for examples. I think we could use `createLocalMethodCall` but that needs a new method name

johanadivare:
> > @JohJohan you were trying to remove partial method call from fluent call. You need to create new parent node or replace root node with new Node instead.
> 
> Okay, i looked into it but i'm affraid i'm not able to fix it. I tried looking at our docs or other code for examples. I think we could use `createLocalMethodCall` but that needs a new method name

@samsonasik maybe you could give another hint on where to find the solution

samsonasik:
@JohJohan you need to replace parent node var with the node, I made a commit for it :) https://github.com/rectorphp/rector-symfony/pull/259/commits/d4d075e18dba8dac10f59c9437f6a35fb51bcfdd

johanadivare:
@samsonasik thanks i will fix the phpstan issues

johanadivare:
@samsonasik i cleanup the commit history by rebasing and create 1 commit for my changes

samsonasik (approved):

src/Rector/Return_/CommandConstantReturnCodeRector.php
```diff
@@ -82,17 +81,16 @@ public function refactor(Node $node): ?Node
             return null;
         }
 
-        if (! $this->nodeNameResolver->isName($node, 'execute')) {
+        if (! $classReflection->hasMethod('execute')) {
```
samsonasik, line 84: This is wrong, hasMethod verify method exists, not check existing method that return parent of it

src/Rector/Return_/CommandConstantReturnCodeRector.php
```diff
@@ -65,11 +64,11 @@ protected function execute(InputInterface $input, OutputInterface $output): int
      */
     public function getNodeTypes(): array
     {
-        return [ClassMethod::class];
+        return [Return_::class];
```
samsonasik, line 67: The old ClassMethod is the correct way, as directly check ClassMethod name

samsonasik (changes requested):
Please revert `CommandConstantReturnCodeRector` to use ClassMethod node, as before, see prev comment

src/Rector/Return_/CommandConstantReturnCodeRector.php
johanadivare: I think i resolved the merge incorrectly at some point thanks for finding it.

johanadivare:
@samsonasik i revert my invalid changes and squashed my commits again

src/Rector/Class_/CommandPropertyToAttributeRector.php
```diff
@@ -118,34 +203,74 @@ public function refactor(Node $node): ?Node
...
+
+            if (! $this->isName($node->name, 'setHidden')) {
+                return null;
+            }
+
+            if (! $node->getArgs()[0]->value instanceof ConstFetch) {
+                return null;
+            }
+
+            $commandHidden = $node->getArgs()[0]
```
samsonasik, line 263: multiple `getArgs()` can be saved to variable, also, if you have `$node->isFirstCallable()` check, you can just use `$node->args`, the `@var Arg $firstArg` is fine after that

johanadivare: I defined a variable for getArgs, i had a look at `$node->isFirstCallable()` but its not a function on `$node`?

johanadivare: Ah its probably `isFirstClassCallable`

johanadivare: Was discussed in https://github.com/rectorphp/rector-symfony/pull/279/files#r1029199016

johanadivare: Its ready now

samsonasik (approved):

samsonasik (changes requested):
One more check on Arg value exist and its value

tests/Rector/Class_/CommandPropertyToAttributeRector/Fixture/commandWithDescriptionAndAliasesAndHiddenAndHelp.php.inc
```diff
@@ -0,0 +1,40 @@
...
+    public static $defaultName = 'sunshine';
+
+    /** @var string|null */
+    public static $defaultDescription = 'Let the sun shine!';
+
+    public function configure(): void
+    {
+        $this
+            ->setAliases(['sun', 'shine'])
+            ->setHidden(false)
```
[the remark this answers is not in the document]
johanadivare: You mean what about `setHidden()` which would set the hidden to `true` right now the code fails so i will add some checks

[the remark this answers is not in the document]
johanadivare: I pushed some code let me know what you think. Also good catch!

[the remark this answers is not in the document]
samsonasik: On 2 conditions:

- no arg (use default value)
- true arg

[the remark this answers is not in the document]
johanadivare: I fixed phpstan issue

tests/Rector/Class_/CommandPropertyToAttributeRector/Fixture/commandWithHidden.php.inc
```diff
@@ -0,0 +1,31 @@
...
+
+class CommandWithHidden extends \Symfony\Component\Console\Command\Command
+{
+    /** @var string|null */
+    public static $defaultName = 'sunshine';
+
+    public function configure(): void
+    {
+        $this
+            ->setHidden();
```
samsonasik, line 13: setHidden(true) test case need to be added, are they need to be removed whatever value pass on it?

johanadivare: Added testcase with `setHidden(true)` false case is handled in https://github.com/rectorphp/rector-symfony/pull/259/files#diff-1ce73e8d15ebe88b93e9a7d208697815fcadb4ab67e8067a0e646e9bee4141db

What do you mean with `are they need to be removed whatever value pass on it?`?

samsonasik: There are 3 possible conditions:

- setHidden()
- setHidden(false)
- setHidden(true)

Are they need to be removed so need check against Arg value ?

johanadivare: I think what you mean is do we need to remove `setHidden` function on all these cases and move them to the `AsCommand` attribute? I would say yes. But not sure that is what you mean

samsonasik: Yes, that's what I mean 👍 , it seems test added now, looks good 👍

samsonasik (approved):

samsonasik (changes requested):
Just general improvement:

1. are the `setHidden()` and `setAliases()` only should be called in `configure()` method, if yes, you may lookup only on `configure()` ClassMethod to improve performance.

```php
$classMethod = $class->getMethod('configure');

if (! $classMethod instanceof \PhpParser\Node\Stmt\ClassMethod) {
    return;
}

// traverse $classMethod->stmts only here...
```

3. After Node changed inside `traverseNodesWithCallable()`, `return \PhpParser\NodeTraverser\NodeTraverser::STOP_TRAVERSAL;` should be called to avoid unnecessary another finding.

You can check its usage in https://github.com/rectorphp/rector-src/blob/a9372bb3e90aa966412b2deab075a44769bb119c/src/NodeAnalyzer/ParamAnalyzer.php

3. If you only need to locate the MethodCall in Expression, you just need to loop the $clasMethod->stmts :

```php
foreach ((array) $classMethod->stmts as $stmt) {
    if (! $stmt instanceof \PhpParser\Node\Stmt\Expression) {
         // stop, possibly there is if else
         return;
    }
}

/** @var \PhpParser\Node\Stmt\Expression $stmts[] */
$stmts =(array) $classMethod->stmts; 
foreach ($stmts as $stmt) {
    if (! $stmt->expr instanceof MethodCall) {
          continue;
     }

     // check setAliases and setHidden() here
}
```

johanadivare:
> 

I did 1/2 but for 3 how do i get the value from the Expression and check that it is SetHidden for example?

samsonasik:
If it always in Expression, you can verify that Expression->expr is MethodCall, then verify after that, but I think that can be next iteration improvement. 

Current implementation looks good 👍

samsonasik (approved):

src/Rector/Class_/CommandPropertyToAttributeRector.php
```diff
@@ -118,34 +206,104 @@ public function refactor(Node $node): ?Node
...
-    private function createAttributeGroup(string $defaultName, ?string $defaultDescription): AttributeGroup
+    private function resolveAliases(Class_ $class): ?Array_
     {
-        $attributeGroup = $this->phpAttributeGroupFactory->createFromClass(self::ATTRIBUTE);
+        $commandAliases = null;
+        $classMethod = $class->getMethod('configure');
+        if (! $classMethod instanceof ClassMethod) {
+            return null;
+        }
+        $this->traverseNodesWithCallable($class->stmts, function (Node $node) use (&$commandAliases) {
```
samsonasik, line 219: Oh, it need to be `$classMethod->stmts`:

```suggestion
        $this->traverseNodesWithCallable((array) $classMethod->stmts, function (Node $node) use (&$commandAliases) {
```

src/Rector/Class_/CommandPropertyToAttributeRector.php
```diff
@@ -118,34 +206,104 @@ public function refactor(Node $node): ?Node
...
-        if (count($property->props) !== 1) {
+        $commandHidden = null;
+        $classMethod = $class->getMethod('configure');
+        if (! $classMethod instanceof ClassMethod) {
             return null;
         }
-        $propertyProperty = $property->props[0];
-        if ($propertyProperty->default instanceof String_) {
-            return $propertyProperty->default->value;
+        $this->traverseNodesWithCallable($class->stmts, function (Node $node) use (&$commandHidden) {
```
samsonasik, line 264: Oh, it need to be `$classMethod->stmts`:

```suggestion
        $this->traverseNodesWithCallable((array) $classMethod->stmts, function (Node $node) use (&$commandHidden) {
```

src/Rector/Class_/CommandDescriptionToPropertyRector.php
```diff
@@ -108,33 +111,42 @@ private function resolveCommandDescription(Class_ $class): ?Node
     private function resolveCommandDescriptionFromSetDescription(Class_ $class): ?Node
     {
         $commandDescription = null;
-
+        $classMethod = $class->getMethod('configure');
+        if (! $classMethod instanceof ClassMethod) {
+            return null;
+        }
         $this->traverseNodesWithCallable($class->stmts, function (Node $node) use (&$commandDescription) {
```
samsonasik, line 118: Oh, it need to be `$classMethod->stmts`:

```suggestion
        $this->traverseNodesWithCallable((array) $classMethod->stmts, function (Node $node) use (&$commandDescription) {
```

samsonasik (changes requested):
Tge traverse to use `(array) $classMethod->stmts` instead of `$class->stmts`

samsonasik:
You can verify it only on Expression with verify:

```php
$expressionStmts = array_filter((array) $classMethod->stmts, fn (\PhpParser\Node\Stmt $stmt): bool => $stmt instanceof \PhpParser\Node\Stmt\Expression);
if (count((array) $classMethod->stmts) > count($expressionStmts)) {

    return;

}
```

so the traverse will only on Expression

src/Rector/Class_/CommandPropertyToAttributeRector.php
johanadivare: ++

src/Rector/Class_/CommandPropertyToAttributeRector.php
johanadivare: ++

src/Rector/Class_/CommandDescriptionToPropertyRector.php
johanadivare: ++

johanadivare:
> ```
> 
>     if (! $stmt instanceof \PhpParser\Node\Stmt\Expression) {
>          // stop, possibly there is if else
>          return;
>     }
> ```

I dont see the benefit yet for checking if it is a `Expression` but it would look something like:
```
 foreach ((array) $classMethod->stmts as $stmt) {
            if (!$stmt instanceof \PhpParser\Node\Stmt\Expression) {
                // stop, possibly there is if else
                continue;
            }

            if (!$stmt->expr instanceof MethodCall) {
                continue;
            }


            if (!$this->isName($stmt->expr->name, 'setAliases')) {
                continue;
            }

            /** @var Arg $arg */
            $arg = $stmt->expr->args[0];
            if (!$arg->value instanceof Array_) {
                return null;
            }

            $commandAliases = $arg->value;

            $parentNode = $stmt->expr->getAttribute(AttributeKey::PARENT_NODE);
            if ($parentNode instanceof MethodCall) {
                $parentNode->var = $stmt->expr->var;
            } else {
                $this->removeNode($stmt->expr);
            }
        }
```
Right?

Or do you still want me to use `traverseNodesWithCallable` but with the check you mentioned so i will only traverse if it is a Expression?

samsonasik:
If you want to loop, you need to double loop, once is for verify if the stmts only Expression with MethodCall Expr, next is to apply changes, directly change will make it already changed while stop in next If statement, which invalid.

Using array_filter() with compare count can be both safer and less code I think.

johanadivare:
> If you want to loop, you need to double loop, once is for verify if the stmts only Expression with MethodCall Expr, next is to apply changes, directly change will make it already changed while stop in next If statement, which invalid.
> 
> Using array_filter() with compare count can be both safer and less code I think, traverse is find, but verify first as it apply change during traversion.

I be honest and dont know 100% sure what is going on but i tried:
```php

        $expressionStmts = array_filter(
            (array)$classMethod->stmts,
            fn(\PhpParser\Node\Stmt $stmt): bool => $stmt instanceof \PhpParser\Node\Stmt\Expression
        );
        if (count((array)$classMethod->stmts) > count($expressionStmts)) {
            return null;
        }
        foreach ((array) $classMethod->stmts as $stmt) {
            if (!$stmt->expr instanceof MethodCall) {
                continue;
            }
            if (!$this->isName($stmt->expr->name, 'setAliases')) {
                continue;
            }

            /** @var Arg $arg */
            $arg = $stmt->expr->args[0];
            if (!$arg->value instanceof Array_) {
                return null;
            }

            $commandAliases = $arg->value;

            $parentNode = $stmt->expr->getAttribute(AttributeKey::PARENT_NODE);
            if ($parentNode instanceof MethodCall) {
                $parentNode->var = $stmt->expr->var;
            } else {
                $this->removeNode($stmt->expr);
            }
        }
```

But this didn't work see test output:
```diff
 
-#[\Symfony\Component\Console\Attribute\AsCommand('sunshine', 'Let the sun shine!', ['sun', 'shine'], false)]
+#[\Symfony\Component\Console\Attribute\AsCommand('sunshine', 'Let the sun shine!', [], false)]
 class CommandWithDescriptionAndAliasesAndHiddenAndHelp extends \Symfony\Component\Console\Command\Command
 {
     public function configure(): void
     {
         $this
+            ->setAliases(['sun', 'shine'])
             ->setHelp('Let the sun shine! Its that simple')
         ;
     }

```

samsonasik (approved):

samsonasik:
Thank you @JohJohan

PR #261, opened 2022-10-25
[RemoveUnusedRequestParamRector] Add rector rule to docs

JohJohan:
@TomasVotruba yesterday I saw we where missing one more rule in the docs. After this it should be complete as the number of rules are same in rector folder and in docs.

I also changed code sample to show the correct output off the rule


TomasVotruba:
Thank you :

PR #263, opened 2022-10-28
[MergeMethodAnnotationToRouteAnnotationRector] Adding single method handling

stefantalen:
While upgrade a legacy project I noticed some `@Method` annotations were left out because the value was a single string.

This update adds support for single methods :)

### DIFF
--- a/src/Rector/ClassMethod/MergeMethodAnnotationToRouteAnnotationRector.php
+++ b/src/Rector/ClassMethod/MergeMethodAnnotationToRouteAnnotationRector.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Rector\Symfony\Rector\ClassMethod;
 
 use PhpParser\Node;
+use PhpParser\Node\Scalar\String_;
 use PhpParser\Node\Stmt\ClassLike;
 use PhpParser\Node\Stmt\ClassMethod;
 use Rector\BetterPhpDocParser\PhpDoc\ArrayItemNode;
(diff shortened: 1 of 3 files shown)


src/Rector/ClassMethod/MergeMethodAnnotationToRouteAnnotationRector.php
```diff
@@ -64,6 +73,13 @@ class DefaultController extends Controller
     public function show($id)
     {
     }
+
+    /**
+     * @Route("/post/{id}", methods={"POST"})
+     */
+    public function post($id)
+    {
+    }
```
TomasVotruba, line 82: Could you keep this sample as small as it was? It is used to generate documentation and it should contain small code samples to make the whole docs easier to read :+1:

tests/Rector/ClassMethod/MergeMethodAnnotationToRouteAnnotationRector/Fixture/fixture.php.inc
```diff
@@ -33,6 +41,13 @@ class DefaultController
     public function show($id)
     {
     }
+
+    /**
+     * @Route("/post/{id}", methods={"POST"})
+     */
+    public function post($id)
+    {
+    }
```
TomasVotruba, line 50: Could you extract the test fixture to a new file? 
So we have this one separated and easier to track in case of regression in the future :+1:

TomasVotruba:
Thank you for the feature :+1: it looks very good :)

TomasVotruba:
Thank you :clap:

Issue #264, opened 2022-11-02
[SYMFONY62] Built-in Cache, Security, Template and Doctrine Attributes


JohJohan:
In the blog from Symfony https://symfony.com/blog/new-in-symfony-6-2-built-in-cache-security-template-and-doctrine-attributes its explained new attributes are introduced to replace the once from SensioFrameworkExtraBundle.

We need a few changes:
Replace attributes `IsGranted` and `Cache`:
```diff
-use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
+use Symfony\Component\Security\Http\Attribute\IsGranted;
-use Sensio\Bundle\FrameworkExtraBundle\Configuration\Cache;
+use Symfony\Component\HttpKernel\Attribute\Cache;
```

Replace `ParamConverter` with `MapEntity` see:
```diff
-use Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter;
+use Symfony\Bridge\Doctrine\Attribute\MapEntity;

 // ...
-#[ParamConverter('post', options: ['mapping' => ['postSlug' => 'slug']])]
-public function addComment(Request $request, Post $post): Response
+public function addComment(
+    Request $request,
+    #[MapEntity(mapping: ['postSlug' => 'slug'])] Post $post
+): Response
 {
     // ...
 }
``` 

I can pick this up.

JohJohan:
See https://github.com/rectorphp/rector-symfony/pull/265

johanadivare:
Only MapEntity has to be done still

stefantalen:
> Only replace `ParamConverter` with `MapEntity` has to be done still

@johanadivare are you working on this? If not, mind if I give it a try?

johanadivare:
@stefantalen feel free i'm working on https://github.com/rectorphp/rector-symfony/pull/279#issuecomment-1322075398

Dont forget to add the rule to symfony62.php :)

stephanvierkant:
```diff
- #[\Sensio\Bundle\FrameworkExtraBundle\Configuration\Cache(maxage: 120, lastModified: 'entity.getUpdatedAt()', etag: 'entity.getETag()')]
+ #[\Symfony\Component\HttpKernel\Attribute\Cache(maxage: 120, lastModified: 'entity.getUpdatedAt()', etag: 'entity.getETag()')]
public function someAction(Entity $entity) : Response
{
}
```

This used to work, but now it gives an error:

[...]

PR #265, opened 2022-11-02
[Symfony62] Replace attributes #264

JohJohan:
Replace `IsGranted`, `Cache` and `Template`

### DIFF
--- a/config/sets/symfony/level/up-to-symfony-62.php
+++ b/config/sets/symfony/level/up-to-symfony-62.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+use Rector\Config\RectorConfig;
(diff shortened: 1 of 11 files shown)


JohJohan:
I didn't do `ParamConverter` as that is a bit more work with the mapping see https://symfony.com/blog/new-in-symfony-6-2-built-in-cache-security-template-and-doctrine-attributes maybe that can be a different ticket?

JohJohan:
If https://github.com/rectorphp/rector-symfony/pull/266 is merged i need to update https://github.com/rectorphp/rector-symfony/blob/a0aecf576c517d721714515d5311f9be555132e1/config/sets/symfony/level/up-to-symfony-62.php#L10

TomasVotruba:
Thank you :clap:

PR #266, opened 2022-11-02
[Symfony61] Add rules for upgrade

JohJohan:
Replace deprecated classes see https://github.com/symfony/symfony/blob/6.1/UPGRADE-6.1.md

TODO https://github.com/symfony/symfony/blob/6.1/UPGRADE-6.1.md#validator still has to be done


JohJohan:
I dont know why https://github.com/rectorphp/rector-symfony/actions/runs/3376828076/jobs/5605018494 fails for test case https://github.com/rectorphp/rector-symfony/blob/1f53ebc9ab27e9e9ef4f89ba03964e6353bbb981/tests/Set/Symfony61/Fixture/replace_deprecated_interfaces_serializer.php.inc it should just skip this as its not part off the Command class

tests/Set/Symfony61/Fixture/set_description_at_top.php.inc
```diff
@@ -0,0 +1,57 @@
...
+<?php
+
+namespace Rector\Symfony\Tests\Set\Symfony61\Fixture;
+
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class SetDescriptionAtTop extends \Symfony\Component\Console\Command\Command
+{
+    protected static $defaultDescription = 'Description';
```
stefantalen, line 39: Why add this test for 6.1? It is just a copy of one of the `CommandDescriptionToPropertyRector` tests

config/sets/symfony/symfony61.php
```diff
@@ -0,0 +1,25 @@
...
+
+use Rector\Config\RectorConfig;
+use Rector\Renaming\Rector\Name\RenameClassRector;
+use Rector\Symfony\Rector\Class_\CommandDescriptionToPropertyRector;
+
+# https://github.com/symfony/symfony/blob/6.1/UPGRADE-6.1.md
+
+return static function (RectorConfig $rectorConfig): void {
+    // @see https://symfony.com/blog/new-in-symfony-5-3-lazy-command-description
+    $rectorConfig->rule(CommandDescriptionToPropertyRector::class);
```
stefantalen, line 13: Since the property is deprecated, why not use the `CommandPropertyToAttributeRector` to change it to an attribute?

JohJohan: Good find! I will change it to `CommandPropertyToAttributeRector`

tests/Set/Symfony61/Fixture/set_description_at_top.php.inc
JohJohan: Okay so i added `CommandPropertyToAttributeRector` should i add a fixture of that test or should this be skipped as this is tested elsewhere?

stefantalen: If you ask me, the expected refactor for this deprecation should be a static property to an attribute.

Using `setName()` and `setDescription()` is not deprecated. The Symfony documentation just prefers it the "new" way for optimizations:

> Defining the $defaultDescription static property instead of using the setDescription() method allows to get the command description without instantiating its class. This makes the php bin/console list command run much faster.

https://symfony.com/doc/current/console.html#configuring-the-command

So my personal expectation for using this set would be a refactors like this:

```diff
-#[\Symfony\Component\Console\Attribute\AsCommand(name: 'some:command']
+#[\Symfony\Component\Console\Attribute\AsCommand(name: 'some:command', description: 'Description')]
class SetDescriptionAtTop extends \Symfony\Component\Console\Command\Command
{
-    protected static $defaultDescription = 'Description';
}
```
and
```diff
+#[\Symfony\Component\Console\Attribute\AsCommand(name: 'some:command', description: 'Description')]
class SetDescriptionAtTop extends \Symfony\Component\Console\Command\Command
{
-    protected static $defaultName = 'some:command';
-    protected static $defaultDescription = 'Description';
}
```
But that last example has already been covered by the tests for `CommandPropertyToAttributeRector`, however we want to test changes using `SymfonySetList::SYMFONY_61`.

Maybe @TomasVotruba can guide us here. Do we need duplicate tests when a Rector is used in a set to verify behaviour of the set?

config/sets/symfony/symfony61.php
```diff
@@ -0,0 +1,24 @@
...
+# https://github.com/symfony/symfony/blob/6.1/UPGRADE-6.1.md
+
+return static function (RectorConfig $rectorConfig): void {
+    $rectorConfig->rule(CommandPropertyToAttributeRector::class);
+
+    $rectorConfig->ruleWithConfiguration(
+        RenameClassRector::class,
+        [
+            // @see https://github.com/symfony/symfony/pull/43982
+            'Symfony\Component\Serializer\Normalizer\ContextAwareDecoderInterface' => 'Symfony\Component\Serializer\Normalizer\DecoderInterface',
```
stefantalen, line 18: Looking at the [UPGRADE-6.1.MD](https://github.com/symfony/symfony/blob/6.1/UPGRADE-6.1.md) I don't see the `ContextAwareDecoderInterface` and `ContextAwareEncoderInterface` being deprecated.

Those are mentioned in the PR you are referring to, but have been reversed in [PR #47150](https://github.com/symfony/symfony/pull/47150)

JohJohan: Also good find thanks!

tests/Set/Symfony61/Fixture/set_description_at_top.php.inc
TomasVotruba: > Maybe @TomasVotruba can guide us here. Do we need duplicate tests when a Rector is used in a set to verify behaviour of the set?


That's not necessary indeed. But the bug should be fixed, so CI pass.

JohJohan: Check i added missing stubs

johanadivare (approved):

samsonasik (approved):

TomasVotruba:
Thank you :+1:

PR #267, opened 2022-11-03
Add failing test fixture for WebTestCaseAssertResponseCodeRector

burkun-sergey:
# Failing Test for WebTestCaseAssertResponseCodeRector

Based on https://getrector.org/demo/6dcc081d-8237-4cc6-8c14-99fdf8cf33a9

PR #268, opened 2022-11-04
[WebTestCaseAssertResponseCodeRector] Removing check for null value

stefantalen:
Fixes https://github.com/rectorphp/rector/issues/7540 after the fixture in https://github.com/rectorphp/rector-symfony/pull/267 has been corrected


TomasVotruba:
Thanks

PR #270, opened 2022-11-08
AddRouteAnnotationRector: add support for multiple controller matching

Haehnchen:
I noticed that only the first "controller" match is migrated for "@Route" annotation. 

So given the pattern (eg "optional placeholder"), will be now also migrated with 2 annotations:

```
YYY:
  path: /unit/{foo}.json
  defaults:
    _controller: XXX::yyyAction

XXX:
  path: /unit/{foo}/{date}.json
  defaults:
    _controller: XXX::yyyAction

```


johanadivare (approved):
Code looks good by my, normaly we would also have a small reproducer also with pull request so we know its fails without your changes or with https://getrector.org/demo

samsonasik:
Please run:

```
```

to make CI green

Haehnchen:
PR updated

samsonasik (approved):

PR #275, opened 2022-11-17
Add failing test for GetParameterToConstructorInjectionRector

mttsch:
The expected behavior would be that the parent constructor is only called once.

### DIFF
--- a/tests/Rector/MethodCall/GetParameterToConstructorInjectionRector/Fixture/two_params_with_parent_constructor.php.inc
+++ b/tests/Rector/MethodCall/GetParameterToConstructorInjectionRector/Fixture/two_params_with_parent_constructor.php.inc
@@ -1,0 +1,55 @@
+<?php
+
+namespace Rector\Tests\Symfony\Rector\MethodCall\GetParameterToConstructorInjectionRector\Fixture;
+
+use Symfony\Bundle\FrameworkBundle\Controller\Controller as SymfonyController;
+
+class Service {
+}
+
+abstract class Controller extends SymfonyController {
+    public function __construct(protected Service $service)
+    {
+        // do something else
+    }
+}
+
+class TestController extends Controller
+{
+    public function test()
+    {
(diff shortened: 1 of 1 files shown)


johanadivare:
That is indeed strange we ca probably handle this in the rector rule

johanadivare:
> @johanadivare I think that I am too unfamiliar with the code to fix this in the near future.

Fair i had a quick look but i dont see where the double constructor is added

mttsch:
@johanadivare I took a quick look when creating this issue and I think it relates to `\Rector\Core\NodeManipulator\ClassDependencyManipulator::addConstructorDependencyWithCustomAssign()` and `\Rector\Core\NodeManipulator\Dependency\DependencyClassMethodDecorator::decorateConstructorWithParentDependencies()`, though I might be mistaken.

mttsch:
@johanadivare I think that I have found the issue and fixed it in rectorphp/rector-src#3094.

mttsch:
@TomasVotruba Should this PR be reopened because it was auto-closed by merging rectorphp/rector-src@9a93704c5aa267807287645f9d07772a554ac960?

TomasVotruba:
It seems this was a bug that was fixed in https://github.com/rectorphp/rector-src/commit/9a93704c5aa267807287645f9d07772a554ac960

If it's not the case, it should be re-opened.

mttsch:
@TomasVotruba Yes, it was fixed by that commit but the test in that commit covers a different rule than this test does so it would cover the change for two different scenarios, which might not be necessary.

I was asking about the reopening because I lack the permission to do so.

TomasVotruba:
I see :) 

In that case, next time the "closes x" should link only issue, not the PR. This test fixture should've been part of the fixing PR so it's complete work in one PR.

mttsch:
> This test fixture should've been part of the fixing PR so it's complete work in one PR.

The fix was in a different repository, thus different PRs.

TomasVotruba:
I see, got it :

PR #276, opened 2022-11-19
Fix bug with privatesCaller not found

PhilippSchreiber:
Fixes https://github.com/rectorphp/rector/issues/7613

### DIFF
--- a/src/Rector/New_/StringToArrayArgumentProcessRector.php
+++ b/src/Rector/New_/StringToArrayArgumentProcessRector.php
@@ -149,7 +149,7 @@ CODE_SAMPLE
      */
     private function splitProcessCommandToItems(string $process): array
     {
-        $privatesCaller = new PrivatesCaller();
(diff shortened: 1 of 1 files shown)


samsonasik (approved):

samsonasik:
@TomasVotruba The `MethodCallRenameWithArrayKey` error seems due to the class removed the `MethodCallRenameWithArrayKey` class on your PR:

-  https://github.com/rectorphp/rector-src/pull/3077

should the class and its usage be reverted ?

samsonasik:
@TomasVotruba this got class `MethodCallRenameWithArrayKey` not found error;

https://github.com/rectorphp/rector-symfony/blob/695c960e7713438ef1077bebcbc6d298a95020d1/config/sets/symfony/symfony41.php#L44

TomasVotruba:
@samsonasik That's intersting... it should fail on rector-src then. We probably need to run PHPStan in CI tests there too.
I'll revert, thanks

TomasVotruba:
Thank you @PhilippSchreiber

PR #277, opened 2022-11-19
[PHPStan] Fix PHPStan notice

samsonasik:
Fix PHPStan notice:

```
composer phpstan
> vendor/bin/phpstan analyse --ansi --error-format symplify

 ! [NOTE] The Xdebug PHP extension is active, but "--xdebug" is not used. This may  
 !        slow down performance and the process will not halt at breakpoints.       

Note: Using configuration file /Users/samsonasik/www/rector-symfony/phpstan.neon.
 394/394 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%


                                                                                    
 [WARNING] Ignored error pattern #Parameter \#1 \$class of method                   
           Rector\\BetterPhpDocParser\\PhpDocInfo\\PhpDocInfo<PHPStan\\PhpDocParser\</>
           Ast\\Node>\:\:(.*?)\(\) expects class\-string, string given# was not     
           matched in reported errors.                  
```

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -22,8 +22,6 @@ parameters:
              message: '#Instead of "instanceof/is_a\(\)" use ReflectionProvider service or "\(new ObjectType\(<desired_type\>\)\)\-\>isSuperTypeOf\(<element_type\>\)" for static reflection to work#'
              path: src/ValueObjectFactory/ServiceMapFactory.php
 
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Thank you

Issue #278, opened 2022-11-21
[SYMFONY62]  Simpler Form Rendering


johanadivare:
In Symfony 6.2 a simpler way is introduced to render forms: https://symfony.com/blog/new-in-symfony-6-2-dx-improvements#simpler-form-rendering

I suggest we create a rule that removes the function call `->createView()` what do you think?

stefantalen:
Good suggestion 👍 , [Symfony docs](https://symfony.com/doc/6.2/forms.html#rendering-forms) don't mention the `->createView()` anymore either.

I would also include a rule to replace the deprecated `AbstractController->renderForm()` with `AbstractController->render()`?

TomasVotruba:
Thanks for suggestion, they both sounds good 👍 
Go for it :)

johanadivare:
Fixed with https://github.com/rectorphp/rector-symfony/pull/279

stollr:
I'll exclude this rule, because it hides the fact, that the template will handle a FormView and not a Form. This might be surprising for newcomers and I like to be explicit about my code ;-)

stefantalen:
The [Symfony documentation](https://symfony.com/doc/6.2/forms.html#rendering-forms) suggest implementing it this way. I see no reason for excluding, other than personal preference 🙂 

Newcomers should read the documentation to avoid surprises 😬

PR #279, opened 2022-11-21
[SYMFONY62] OhMySMTP, ArgumentValueResolverInterface, Security etc. #278

JohJohan:


johanadivare:
Fixes https://github.com/rectorphp/rector-symfony/issues/278

johanadivare:
What is the best way to remove the function call to `->createView()`? I can make a rule in class method and check if it has a render function call that contains `form`?

stefantalen:
1. Check arguments of `AbstractController::render()`
2. Check if argument is an instance of `FormInstance`
3. Remove `->createView()` from argument

However, doing it this way might not update implementations like:
```php
$form = $contactForm->createView();
return $this->render('template.html.twig', [
    contact_form => $form,
];
```
But I think it'll be safer to limit this feature to update calls within the `render()` method only.

johanadivare:
@stefantalen i agree, and the implementation like you mention also still works

TomasVotruba:
Let me know when this is ready.

The createView() removal will indeed need a custom rule.  The conditions list seem valid.

johanadivare (approved):

samsonasik:
Please run:

```
```

to make CI green.

src/Rector/MethodCall/SimplifyFormRenderingRector.php
```diff
@@ -0,0 +1,104 @@
...
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if (! $this->isObjectType(
+            $node->var,
+            new ObjectType('Symfony\Bundle\FrameworkBundle\Controller\AbstractController')
+        )) {
+            return null;
+        }
+
```
samsonasik, line 74: first class callable should be skipped:

```suggestion
        if ($node->isFirstClassCallable()) {
            return null;
        }
```

johanadivare: ++

src/Rector/MethodCall/SimplifyFormRenderingRector.php
```diff
@@ -0,0 +1,117 @@
...
+            }
+            $replaced = true;
+            $arrayItem->value = $arrayItem->value->var;
+        }
+
+        if (! $replaced) {
+            return null;
+        }
+
+        return $node;
```
samsonasik, line 115: You can move this logic to separate method to make PHPStan green, eg: `

```php
/**
 * @param ArrayItem[]|null[] $items
 */
private function processRemoveCreateView(array $items): ?MethodCall
```

samsonasik:
You can verify phpstan with run:

```
composer phpstan
```

src/Rector/MethodCall/SimplifyFormRenderingRector.php
```diff
@@ -0,0 +1,117 @@
...
+            return null;
+        }
+
+        if (! isset($node->args[1])) {
+            return null;
+        }
+
+        if (! $node->args[1] instanceof Arg) {
+            return null;
+        }
```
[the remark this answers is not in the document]
johanadivare: I can check isset on value?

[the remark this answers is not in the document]
samsonasik: I see, assign to variable with `@var Arg $variableName` should be ok

src/Rector/MethodCall/SimplifyFormRenderingRector.php
johanadivare: ++

src/Rector/MethodCall/SimplifyFormRenderingRector.php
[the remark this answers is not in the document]
johanadivare: ++

johanadivare:
@samsonasik its ready now :)

johanadivare:
@TomasVotruba its ready

johanadivare:
Note i added a bit more rules

config/sets/symfony/symfony62.php
```diff
@@ -16,6 +23,53 @@
...
+            // @see https://github.com/symfony/symfony/pull/45034
+            new MethodCallRename(
+                'Symfony\Component\HttpFoundation\Request',
+                'getContentType',
+                'getContentTypeFormat'
+            ),
+            // @see https://github.com/symfony/symfony/pull/47711
+            new MethodCallRename('Symfony\Component\Mime\Email', 'attachPart', 'addPart'),
+            // @see https://github.com/symfony/symfony/pull/47363
+            new MethodCallRename(
```
johanadivare, line 60: Is this allowed because return type is diffrent supports: bool resolve: array

TomasVotruba:
Thank you :clap:

config/sets/symfony/symfony62.php
samsonasik: Per comment on https://github.com/rectorphp/rector-symfony/pull/283#discussion_r1029286911

this part may  need to be reverted

PR #283, opened 2022-11-22
[SYMFONY53] Replace deprecated functions

JohJohan:


config/sets/symfony/symfony53.php
```diff
@@ -59,6 +59,17 @@
...
+        // @see https://github.com/symfony/symfony/pull/39020
+        new MethodCallRename(
+            'Symfony\Component\PropertyInfo\Type',
+            'getCollectionKeyType',
+            'getCollectionKeyTypes'
+        ),
+        new MethodCallRename(
+            'Symfony\Component\PropertyInfo\Type',
+            'getCollectionValueType',
+            'getCollectionValueTypes'
```
johanadivare, line 71: Is it okay to replace these functions because it gets a different return type which probably needs manual change

TomasVotruba: This would have to be handled with custom rule, as this bare change would break existing code.

johanadivare: Okay i will create a issue and close this pull request

PR #291, opened 2022-11-25
[SYMFONY62] Replace ParamConverter with MapEntity

stefantalen:
See https://github.com/rectorphp/rector-symfony/issues/264

This is my first time working with Rector and the AST so please add feedback 🙂 

Right now I'm confused why the `use` is not being removed (I thought `RenamedClassesDataCollector` would take care of that) and why the `MapEntity` is placed on a new line instead of prepended to the parameter.

### To do:
- [x] ~Set `MapEntity` on the same line as the class method parameter~ [Rector convention](https://github.com/rectorphp/rector-symfony/pull/291#discussion_r1032600236)
- [x] Remove `use` of `ParamConverter`
- [x] Add fixture with multiple ParamConverters
- [x] Add fixture(s) that should be skipped
- [x] Add fixture with other options (`id`, `exclude`, `strip_null`, `entity_manager` and `evict_cache`)
- [x] Fix CI

### Discussion:
1. Should this also replace `Entity` attributes since that is a shortcut for using `expr`?

### DIFF
--- a/config/sets/symfony/symfony62.php
+++ b/config/sets/symfony/symfony62.php
@@ -8,6 +8,7 @@ use Rector\Renaming\Rector\MethodCall\RenameMethodRector;
 use Rector\Renaming\Rector\Name\RenameClassRector;
 use Rector\Renaming\ValueObject\MethodCallRename;
 use Rector\Renaming\ValueObject\RenameClassAndConstFetch;
+use Rector\Symfony\Rector\ClassMethod\ParamConverterAttributeToMapEntityAttributeRector;
 use Rector\Symfony\Rector\MethodCall\SimplifyFormRenderingRector;
 
 return static function (RectorConfig $rectorConfig): void {
--- a/config/sets/symfony/symfony62.php
+++ b/config/sets/symfony/symfony62.php
@@ -87,4 +87,6 @@ return static function (RectorConfig $rectorConfig): void {
             'MAX_USERNAME_LENGTH'
         ),
     ]);
+
+    $rectorConfig->rule(ParamConverterAttributeToMapEntityAttributeRector::class);
 };
--- a/src/Rector/ClassMethod/ParamConverterAttributeToMapEntityAttributeRector.php
+++ b/src/Rector/ClassMethod/ParamConverterAttributeToMapEntityAttributeRector.php
@@ -12,10 +12,10 @@ use PhpParser\Node\Expr\Array_;
 use PhpParser\Node\Expr\ArrayItem;
 use PhpParser\Node\Expr\Variable;
 use PhpParser\Node\Identifier;
+use PhpParser\Node\Name\FullyQualified;
 use PhpParser\Node\Scalar\String_;
 use PhpParser\Node\Stmt\ClassMethod;
 use PhpParser\Node\Stmt\Use_;
-use Rector\Core\Configuration\RenamedClassesDataCollector;
 use Rector\Core\Rector\AbstractRector;
 use Rector\Core\ValueObject\PhpVersionFeature;
 use Rector\NodeTypeResolver\Node\AttributeKey;
@@ -35,9 +35,10 @@ final class ParamConverterAttributeToMapEntityAttributeRector extends AbstractRe
 
     private const MAP_ENTITY_CLASS = 'Symfony\Bridge\Doctrine\Attribute\MapEntity';
 
+    private ?Use_ $useNode;
+
     public function __construct(
         private readonly PhpAttributeAnalyzer $phpAttributeAnalyzer,
-        private readonly RenamedClassesDataCollector $renamedClassesDataCollector,
     ) {
     }
 
@@ -93,7 +94,7 @@ CODE_SAMPLE
     public function refactor(Node $node): ?Node
     {
         if (($node instanceof Use_) && $this->isName($node, self::PARAM_CONVERTER_CLASS)) {
-            $this->removeNode($node);
+            $this->useNode = $node;
         }
 
         if (
@@ -104,7 +105,12 @@ CODE_SAMPLE
             return null;
         }
 
-        return $this->refactorParamConverter($node);
+        $this->refactorParamConverter($node);
+
+        if ($this->useNode) {
+            $this->removeNode($this->useNode);
+        }
+        return $node;
     }
 
     private function refactorParamConverter(ClassMethod $classMethod): Node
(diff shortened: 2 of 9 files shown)


tests/Rector/ClassMethod/ParamConverterAttributeToMapEntityAttributeRector/Fixture/replace_paramconverter.php.inc
```diff
@@ -0,0 +1,45 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Annotation\Route;
+
+class ReplaceParamConverterWithMapping
+{
+    #[Route('/blog/{blogSlug}', name: 'blog_detail')]
+    public function convert(
+        Request $request,
+        #[\Symfony\Bridge\Doctrine\Attribute\MapEntity(mapping: ['blogSlug' => 'slug'])] stdClass $blog
```
samsonasik, line 38: it needs separate line as reported in failing test:

```suggestion
        #[\Symfony\Bridge\Doctrine\Attribute\MapEntity(mapping: ['blogSlug' => 'slug'])]
        stdClass $blog,
```

stefantalen: So you are saying it is impossible to have the attribute on the same line as the parameter? Changing the test feels like cheating 😅

samsonasik: That's Rector convension to not inline it https://github.com/rectorphp/rector-src/blob/8ef808705cd8f467602f2411ec831552648295ae/src/PhpParser/Printer/BetterStandardPrinter.php#L469

config/sets/symfony/symfony62.php
```diff
@@ -86,4 +87,6 @@
             'MAX_USERNAME_LENGTH'
         ),
     ]);
+
+    $rectorConfig->rule(ParamConverterAttributeToMapEntityAttributeRector::class);
```
johanadivare, line 91: @samsonasik there was some sort of guideline about setting rules was it specific rules like these should be done last? If so should we also move `SimplifyFormRenderingRector`  arround here?

src/Rector/ClassMethod/ParamConverterAttributeToMapEntityAttributeRector.php
```diff
@@ -0,0 +1,163 @@
...
+    {
+        foreach ($classMethod->attrGroups as $attrGroup) {
+            foreach ($attrGroup->attrs as $attr) {
+                if ($this->isName($attr, self::PARAM_CONVERTER_CLASS)) {
+                    $foundAttribute[] = $attr;
+                }
+            }
+        }
+
+        foreach ($foundAttribute as $attr) {
```
johanadivare, line 115: `$foundAttribute` could in theory be undefined if you dont check
```
     if (! $this->phpAttributeAnalyzer->hasPhpAttribute($node, self::PARAM_CONVERTER_CLASS)) {
            return null;
        }
```

that is why you get phpstan error: `  - '#Implicit array creation is not allowed \- variable \$foundAttribute might not exist#'`

stefantalen:
@TomasVotruba should I make a change to get the checks passed or is this an action issue?

TomasVotruba:
That's a report from Rector that should be fixed, as we cannot contribute to your code base.

stefantalen:
@TomasVotruba Makes sense, didn't check the actual scripts that run in the failing action 😅

Looks like it is ready to go 🚀

src/Rector/ClassMethod/ParamConverterAttributeToMapEntityAttributeRector.php
```diff
@@ -0,0 +1,197 @@
...
+            ]
+        );
+    }
+
+    /**
+     * @return array<class-string<Node>>
+     */
+    public function getNodeTypes(): array
+    {
+        return [Use_::class, ClassMethod::class];
```
TomasVotruba, line 91: This rule should only handle the `ClassMethod`. What is the `Use_` here for?

stefantalen: It removes the `use Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter;` when it is no longer needed.

Or should a rule not be responsible for cleaning unused _use_ statements?

TomasVotruba: > Or should a rule not be responsible for cleaning unused use statements?

I see. That's handled somewhere in post-rectors, so you don't have to worry about it in every rule.

See https://github.com/rectorphp/rector-src/pull/3358

TomasVotruba:
To fix CI, there are 2 commands to run - a coding standard:

```bash
composer fix-cs
```

To update tests:

```bash
UT=1 vendor/bin/phpunit
```

src/Rector/ClassMethod/ParamConverterAttributeToMapEntityAttributeRector.php
stefantalen: I deleted the logic for the removal, however my tests now fail because the _use_ is still there.

Do I update the tests and leave them in?

TomasVotruba (approved):

TomasVotruba:
Thank you, let's ship this to test it :)

stephanvierkant:
This doesn't work when using named arguments:

    #[ParamConverter('document', class: Document::class, options: ['mapping' => ['document' => 'id']])]

This codes stays untouched, because `$attribute->args[1]->name->name !== 'options'` (from \Rector\Symfony\Rector\ClassMethod\ParamConverterAttributeToMapEntityAttributeRector::refactorAttribute)

johanadivare:
> This doesn't work when using named arguments:
> 
> ```
> #[ParamConverter('document', class: Document::class, options: ['mapping' => ['document' => 'id']])]
> ```
> 
> This codes stays untouched, because `$attribute->args[1]->name->name !== 'options'` (from \Rector\Symfony\Rector\ClassMethod\ParamConverterAttributeToMapEntityAttributeRector::refactorAttribute)

Good you share example, i think its nice if you add pull request with expected result. That it is probably converted to MapEntity. Then someone can pick it up to create a fix?

stefantalen:
> This doesn't work when using named arguments:

I'm not surprised, `ParamConverter` has some special sauce so I had to make some shortcuts to try and figure out if it is used for Doctrine :) 

I'll create a new PR with your example and try to find a more robust solution

Issue #295, opened 2022-11-25
CommandConstantReturnCodeRector does not add the `use` if needed.


VincentLanglet:
cc https://github.com/rectorphp/rector-symfony/pull/256 (@JohJohan)

Let's say I have a BaseCommand which extends Command.

Then the change from CommandConstantReturnCodeRector is
```
 class SomeCommand extends BaseCommand
 {
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-         return 0;
+         return Command::SUCCESS;
     }
 }
```
And this is breaking my code.

So the `use Symfony\Component\Console\Command\Command;` should be added too.

But maybe it's even better to propose the following change:
```
 class SomeCommand extends BaseCommand
 {
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-         return 0;
+         return self::SUCCESS;
     }
 }
```
?

Issue #293, opened 2022-11-25
Rector is adding useless cast (int) to Command::Success


VincentLanglet:
Hi,

Now https://github.com/rectorphp/rector-symfony/issues/255 is implemented (cc @johanadivare), there is an issue.
I would say it's the rule https://github.com/rectorphp/rector-symfony/blob/main/docs/rector_rules_overview.md#consoleexecutereturnintrector

When running, rector with the `SymfonyLevelSetList::UP_TO_SYMFONY_61`
I get
```
return 0;
```
changed to
```
 return (int) Command::SUCCESS;
```

If I try to change to
```
 return Command::SUCCESS;
```
Running rector give me again
```
 return (int) Command::SUCCESS;
```

I saw you just did a PR https://github.com/rectorphp/rector-symfony/pull/290 @TomasVotruba, dont know if it was for this.

stefantalen:
Looks like a combination of rules, since the [tests](https://github.com/rectorphp/rector-symfony/pull/256/files#diff-6dcce22fe7445f6cfe9680ee4592d57d0d118376b5009b6a9e2cc1ca7f4c864e) are valid and the expected result doesn't show the cast 🙂

VincentLanglet:
I would say it might be a bug with `ConsoleExecuteReturnIntRector` with cast const to int, even if they are already int.

stefantalen:
I think somewhere in between the rules: 
- [`ConsoleExecuteReturnIntRector` return 0](https://getrector.org/demo/4c942cb5-dec7-4546-9b6f-43a4c8ee7c88) 🎉 
- [`ConsoleExecuteReturnIntRector` return constant](https://getrector.org/demo/0e736e5e-3e6a-4ff8-85e5-7f4a56df44b8) 🎉 
- [`CommandConstantReturnCodeRector` return 0](https://getrector.org/demo/31fc4d3c-8223-4f13-a50a-46ef7802984b) 🎉 
- [Combining rules, return 0](https://getrector.org/demo/d80a473b-9201-46a1-bb32-5844bd36927e) 🔥 

cc @JohJohan

samsonasik:
combination between `ConsoleExecuteReturnIntRector` and `CommandConstantReturnCodeRector` seems cause it https://getrector.org/demo/d80a473b-9201-46a1-bb32-5844bd36927e

VincentLanglet:
[...]

PR #296, opened 2022-11-25
Fix issues with ConsoleExecuteReturnIntRector and CommandConstantReturnCodeRector

VincentLanglet:
There is two issues, reported by 
Closes https://github.com/rectorphp/rector-symfony/issues/295
Closes https://github.com/rectorphp/rector-symfony/issues/293

The CommandConstantReturnCodeRector is changing `0` to `Command::Success`, even if `Command` is not imported.
This is breaking the code

Then ConsoleExecuteReturnIntRector is adding `(int)` before `Command::Success` before when resolving with phpstan the constant, an ErrorType is found.

This PR fix both of the issues.

### DIFF
--- a/src/Rector/ClassMethod/CommandConstantReturnCodeRector.php
+++ b/src/Rector/ClassMethod/CommandConstantReturnCodeRector.php
@@ -50,7 +50,7 @@ class SomeCommand extends Command
 {
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        return Command::SUCCESS;
+        return Symfony\Component\Console\Command\Command::SUCCESS;
     }
 
 }
@@ -118,7 +118,7 @@ CODE_SAMPLE
         }
 
         return $this->nodeFactory->createShortClassConstFetch(
-            'Command',
+            'Symfony\Component\Console\Command\Command',
             SymfonyCommandConstantMap::RETURN_TO_CONST[$lNumber->value]
         );
     }
--- a/src/Rector/ClassMethod/ConsoleExecuteReturnIntRector.php
+++ b/src/Rector/ClassMethod/ConsoleExecuteReturnIntRector.php
@@ -16,6 +16,7 @@ use PhpParser\Node\Stmt\Class_;
 use PhpParser\Node\Stmt\ClassMethod;
 use PhpParser\Node\Stmt\Return_;
 use PhpParser\NodeTraverser;
+use PHPStan\Type\ErrorType;
 use PHPStan\Type\IntegerType;
 use PHPStan\Type\ObjectType;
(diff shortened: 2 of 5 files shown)


src/Rector/ClassMethod/ConsoleExecuteReturnIntRector.php
```diff
@@ -229,6 +230,9 @@ private function setReturnTo0InsteadOfNull(Return_ $return): void
         }
 
         $staticType = $this->getType($return->expr);
+        if ($staticType instanceof ErrorType) {
+            return;
+        }
```
samsonasik, line 235: this can be removed as now uses correct constant

VincentLanglet: Does it make sens to cast to int an unknown constant ?

I would have say it was ok to keep this check.

samsonasik (approved):

src/Rector/ClassMethod/ConsoleExecuteReturnIntRector.php
samsonasik: Oh, it seems you removed the ErrorType check 👍  https://github.com/rectorphp/rector-symfony/pull/296/commits/b4ede768547e400d131f66219c4bf2ac25749b83, we can verify in the future if it actually needed or not to verify ErrorType :)

samsonasik:
Thank you @VincentLanglet

src/Rector/ClassMethod/ConsoleExecuteReturnIntRector.php
VincentLanglet: Yes, we will need to take a look at https://getrector.org/demo/5359325c-f447-4c15-b703-31ba2ec5ddee when the sandbox will be updated

samsonasik: that will be updated in next 4 hours.

src/Rector/ClassMethod/CommandConstantReturnCodeRector.php
```diff
@@ -118,7 +118,7 @@ private function convertNumberToConstant(LNumber $lNumber): ?ClassConstFetch
         }
 
         return $this->nodeFactory->createShortClassConstFetch(
```
samsonasik, line 120: Oops, it should be `createClassConstFetch` instead of `createShortClassConstFetch`, I will make a patch :)

samsonasik: https://github.com/rectorphp/rector-symfony/pull/297

johanadivare:
Cool @VincentLanglet thanks for the fix!

PR #302, opened 2022-11-28
Correctly refector argument `message` and support static func

JohJohan:
Refactor param message and also refector static asserts


johanadivare:
Based on https://github.com/rectorphp/rector-symfony/pull/300 i improved other rules to also refactor `message` param and add support for static function calls for asserts

TomasVotruba:
Thank you 🙏

PR #303, opened 2022-12-04
Add redirect response return support to ResponseReturnTypeControllerActionRector

TomasVotruba:


johanadivare:
@TomasVotruba should we also add support for:
`JsonResponse` when we use `json` 
`BinaryFileResponse` when we use `file`
`StreamedResponse` when we use `stream`

TomasVotruba:
@johanadivare Definitelly, that would be very useful :+1:  Could you handle that?

johanadivare:
> @johanadivare Definitelly, that would be very useful +1 Could you handle that?

Will do :)

TomasVotruba:
@johanadivare Thank you :bow:

johanadivare:
See https://github.com/rectorphp/rector-symfony/pull/306

PR #309, opened 2022-12-05
[ResponseReturnTypeControllerActionRector] function is from this class

JohJohan:


samsonasik (approved):

TomasVotruba:
Could you add a test case so we have this covered?

johanadivare:
As you can see i added `isResponseReturn` otherwise it would still add `: ...\Response`

TomasVotruba:
Thank you :+1:

PR #305, opened 2022-12-05
Add failing test fixture for `CommandPropertyToAttributeRector`

OskarStark:
# Failing Test for CommandPropertyToAttributeRector

Based on https://getrector.org/demo/144bbc1f-e170-45a9-b72f-d8daa0686db0

PR #312, opened 2022-12-07
[CommandPropertyToAttributeRector] Also get other attribute values (fix #305)

JohJohan:


johanadivare:
The [unit](https://github.com/rectorphp/rector-symfony/actions/runs/3637517187/jobs/6138616339) test fails because argument `key` are removed, should we add these keys by default what is the best practice for this?

[phpstan](https://github.com/rectorphp/rector-symfony/actions/runs/3637517227/jobs/6138616959)
 fails because ```  src/Rector/Class_/CommandPropertyToAttributeRector.php:35
 ------------------------------------------------------------------------
  - '#Class cognitive complexity is 63, keep it under 50#'``` any idea how i can make this less complex?

tests/Rector/Class_/CommandPropertyToAttributeRector/Fixture/commandWithAttributeNameSetButDescriptionNot.php.inc
```diff
@@ -0,0 +1,24 @@
+<?php
```
johanadivare, line 1: Should fixture file names be named `like_this.php.inc` or `likeThis.php.inc`?

samsonasik: `like_this.php.inc`

OskarStark:
Thank you very much!

tests/Rector/Class_/CommandPropertyToAttributeRector/Fixture/commandWithAttributeNameSetButDescriptionNot.php.inc
johanadivare: I updated the test fixture file names for CommandPropertyToAttributeRector

OskarStark:
Any chances to get this shipped in the next release?

TomasVotruba:
> The [unit](https://github.com/rectorphp/rector-symfony/actions/runs/3637517187/jobs/6138616339) test fails because argument key are removed,

The keys should be kept. The code should not be changed at all and `refactor()` should return `null`.

> any idea how i can make this less complex?

This means a service should be extracted

johanadivare:
> Any chances to get this shipped in the next release?

Not sure when i'm able to finish this up, i only need to make the class less complex still

samsonasik:
You can create another service, and inject into construct method, then use the service

johanadivare:
> You can create another service, and inject into construct method, then use the service

I created a CommandHelper class to move some functions let me know if that is okay

samsonasik (approved):
Looks good

src/Helper/CommandHelper.php
```diff
@@ -0,0 +1,201 @@
...
+use Rector\NodeNameResolver\NodeNameResolver;
+use Rector\NodeRemoval\NodeRemover;
+use Rector\NodeTypeResolver\Node\AttributeKey;
+use Rector\NodeTypeResolver\NodeTypeResolver;
+use Rector\Php80\NodeAnalyzer\PhpAttributeAnalyzer;
+use Rector\PhpDocParser\NodeTraverser\SimpleCallableNodeTraverser;
+
+/**
+ * @see \Rector\Symfony\Tests\Rector\ClassMethod\TemplateAnnotationToThisRenderRector\TemplateAnnotationToThisRenderRectorTest
+ */
```
samsonasik, line 27: The `@see` seems not needed as pointed to different test

johanadivare: Good find i copy paste it from other class and forgot this.

samsonasik:
Let's merge it so we have faster feedback 👍

samsonasik:
Thank you @JohJohan

PR #313, opened 2022-12-07
Add failing test fixture for `CommandPropertyToAttributeRector`

mvhirsch:
# Failing Test for CommandPropertyToAttributeRector

Based on https://getrector.org/demo/19d05cac-0ca7-423f-9e5e-08483d221bd2

I'm adding descriptions to my commands, but rector keeps removing them - even if there is a [test stating skip if attributes set](https://github.com/JohJohan/rector-symfony/blob/main/tests/Rector/Class_/CommandPropertyToAttributeRector/Fixture/skip_command_with_attributes_set.php.inc). I guess this is just a missing behavior.

### DIFF
--- a/tests/Rector/Class_/CommandPropertyToAttributeRector/Fixture/skip_command_with_description_set.php.inc
+++ b/tests/Rector/Class_/CommandPropertyToAttributeRector/Fixture/skip_command_with_description_set.php.inc
@@ -1,0 +1,12 @@
+<?php
+
+namespace Rector\Symfony\Tests\Rector\Class_\CommandPropertyToAttributeRector\Fixture;
+
+#[\Symfony\Component\Console\Attribute\AsCommand('frozen', 'let it go')]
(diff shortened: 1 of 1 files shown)


mvhirsch:
Related to https://github.com/rectorphp/rector-symfony/pull/305.

johanadivare:
Will be fixed by https://github.com/rectorphp/rector-symfony/pull/312, also is your test fixture a valid command as the AsCommand attribute is above the namespace? And rector test fails because off that

mvhirsch:
Not sure why there is a doubled namespace in there. I copied the [original file](https://github.com/rectorphp/rector-symfony/blob/f77e49313ef90366441b9c0bb38943c14cdda46d/tests/Rector/Class_/CommandPropertyToAttributeRector/Fixture/skip_command_with_attributes_set.php.inc) since it's already there, reproducing my issue (partly). I think this should be merged. The filename states "skip command with attribute_**s**_", but only tests "name".

What do you think?

mvhirsch:
After looking at #312, I think this is already handled. So nevermind.

PR #314, opened 2022-12-07
[WebTestCaseAssertResponseCodeRector] Correctly handle setting null

JohJohan:
Fixes #274

PR #315, opened 2022-12-07
Handle sprintf custom messages for asserts

JohJohan:
Todo handle all phpunit asserts that can have 'sprintf' function

### DIFF
--- a/src/Rector/MethodCall/WebTestCaseAssertIsSuccessfulRector.php
+++ b/src/Rector/MethodCall/WebTestCaseAssertIsSuccessfulRector.php
@@ -6,9 +6,9 @@ namespace Rector\Symfony\Rector\MethodCall;
 
 use PhpParser\Node;
 use PhpParser\Node\Arg;
+use PhpParser\Node\Expr\FuncCall;
 use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\StaticCall;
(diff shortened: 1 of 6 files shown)


johanadivare:
We should handle `sprintf` when we refactor assertX(x sprintf('message'));

More info: https://github.com/rectorphp/rector-symfony/pull/314#issuecomment-1340810137 & https://github.com/rectorphp/rector-symfony/pull/314#issuecomment-1340859044

samsonasik:
You may can use NodeTransformer to get extracted data, then next combine them as string value https://github.com/rectorphp/rector-src/blob/f35326d7010fa42458b44d3896b4791dcc22f78a/src/PhpParser/NodeTransformer.php

johanadivare:
@samsonasik i added a check if function call and leave the argument as it is is that okay?

johanadivare:
When i have latest composer package updates and run `vendor/bin/rector` i do not get any changes not sure why https://github.com/rectorphp/rector-symfony/actions/runs/3675860795/jobs/6215800640 is failing then

samsonasik:
should be green after PR https://github.com/rectorphp/rector-symfony/pull/317 merged and you can rebase.

samsonasik (approved):

TomasVotruba:
Thank you :clap:

PR #321, opened 2022-12-21
Fix class name for Symfony 6.2

fabpot:
This class has been moved in https://github.com/symfony/symfony/pull/47760


samsonasik:
Thank you @fabpot

Issue #327, opened 2023-01-19
[SYMFONFY33] ArgumentAdderRector is configured incorrectly


Tiefschwarz:
# Bug Report

## Minimal PHP Code Causing Issue

See https://getrector.com/demo/fec44916-df27-4cd0-ab29-f8422142516f (I copied the rector configuration from the `config/sets/symfony/symfony33.php` that seems to cause the issue)

```php
<?php

use Symfony\Component\DependencyInjection\ContainerBuilder;

final class DemoFile
{
    public function run()
    {
        $container = new ContainerBuilder();
        $container->compile();
    }
}
```

Minimal configuration necessary;

```php
<?php

use Rector\Config\RectorConfig;
use Rector\Symfony\Set\SymfonySetList;

return static function (RectorConfig $rectorConfig): void {
    $rectorConfig->sets([
        SymfonySetList::SYMFONY_33,
    ]);
};
```

## Expected Behavior

Rector changes the file to:

```php
<?php

use Symfony\Component\DependencyInjection\ContainerBuilder;

final class DemoFile
{
    public function run()
    {
        $container = new ContainerBuilder();
        $container->compile(false);
    }
}
```

The expected behavior is based on the fact that Symfony throws a deprecation error when the `compile` method is used without an argument.

For reference:

- https://github.com/symfony/symfony/blob/v3.3.18/src/Symfony/Component/DependencyInjection/ContainerBuilder.php#L709
- https://github.com/symfony/symfony/blob/v4.0.0/src/Symfony/Component/DependencyInjection/ContainerBuilder.php#L684

## Actual Behavior

Rector returns the following error:

```
 "System error: "Previous position does not has default value"

         Stack trace:
         #0 vendor/rector/rector/rules/Arguments/Rector/ClassMethod/ArgumentAdderRector.php(182):
         Rector\Arguments\Rector\ClassMethod\ArgumentAdderRector->fillGapBetweenWithDefaultValue(Object(PhpParser\Node\
         Expr\MethodCall), 2)
         #1 vendor/rector/rector/rules/Arguments/Rector/ClassMethod/ArgumentAdderRector.php(171):
         Rector\Arguments\Rector\ClassMethod\ArgumentAdderRector->processMethodCall(Object(PhpParser\Node\Expr\MethodCa
         ll), 0, 2)
         #2 vendor/rector/rector/rules/Arguments/Rector/ClassMethod/ArgumentAdderRector.php(120):
[...]

PR #328, opened 2023-01-19
[SYMFONY33] Fix configuration for the ArgumentAdderRector of the compile method of the ContainerBuilder (#327)

Tiefschwarz:
* Corrected the position of the argument that should be added
* Corrected the default value of the argument
* Set the name of the argument that is added
* Added unit test

### DIFF
--- a/config/sets/symfony/symfony33.php
+++ b/config/sets/symfony/symfony33.php
@@ -13,7 +13,7 @@ use Rector\Symfony\Rector\ClassConstFetch\ConsoleExceptionToErrorEventConstantRe
 
 return static function (RectorConfig $rectorConfig): void {
     $rectorConfig->ruleWithConfiguration(ArgumentAdderRector::class, [
-        new ArgumentAdder('Symfony\Component\DependencyInjection\ContainerBuilder', 'compile', 2, '__unknown__', 0),
+        new ArgumentAdder('Symfony\Component\DependencyInjection\ContainerBuilder', 'compile', 0, 'resolveEnvPlaceholders', false),
         new ArgumentAdder(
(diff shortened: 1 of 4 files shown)


TomasVotruba:
Thank you :+1:

rene-bos:
Hi @TomasVotruba, at [driftingly/rector-laravel](https://github.com/driftingly/rector-laravel) we have the same problem with the Rector job as this PR (`Error: 'fix/containerbuilder-compile-argument-adder-rector-configuration' branch not found.`). This can be seen in [this job run](https://github.com/driftingly/rector-laravel/actions/runs/4116130795/jobs/7105769970).

We use the same GitHub Actions configuration as here. Do you have a solution for this already? Maybe on some of the other Rector extension repositories? It is a bit diffucult for contributors to spot that they have to run both `vendor/bin/rector process` and after that `composer run fix-cs` when the Rector jobs fails on the `Run EndBug/add-and-commit` step..

TomasVotruba:
Hi, we actually run Rector only for the core contributors. For the rest there is an issue with access as they're external contributors :)

This is the workflow we use on packages: https://github.com/rectorphp/reusable-workflows/blob/main/.github/workflows/rector.yaml

TomasVotruba:
Here is the 2nd piece of puzzle, that checks the ownership: https://github.com/rectorphp/rector-symfony/blob/89e994ffb68921d571bbf5b8e1d59280b7f30a50/.github/workflows/code_analysis.yaml#L9-L18

TomasVotruba:
For your case, add this workflow locally and tune it as needed. The reusable-workflows repository can change any time and is only for our internal packages.

PR #333, opened 2023-01-22
[Symfony26] Fix RedirectToRouteRector with non-default reference type

ratza:
When the code passes a non-default `$referenceType` to the `generateUrl()` call the RedirectToRouteRector should not change the type of the resulting URL. That would be a breaking change.

The PR takes into account the default value defined in the user's UrlGeneratorInterface. But that's also the part I'm not so sure about - couldn't find a similar example in the existing code in reasonable time.

First time digging into the code of the tool, so any suggestions to improve are welcome.


TomasVotruba:
Thanks 🙏 9

PR #340, opened 2023-02-02
Remove check-markdown command from composer fix-cs and composer docs

samsonasik:
Per discussion on https://github.com/rectorphp/rector/issues/7752#issuecomment-1413397028


samsonasik:

PR #348, opened 2023-02-08
[FRAMEWORK_EXTRA_61] Replacing rules with existing set

stefantalen:
Came across some duplicate rules, made more sense to use the other set


TomasVotruba:
Neat :+1:

Issue #352, opened 2023-02-10
Symfony DataMapperInterface iterable vs Traversable


alleknalle:
Hi there,

Since Symfony 5.3 the `Symfony\Component\Form\DataMapperInterface` uses `\Traversable` instead of `iterable` for `mapFormsToData` and `mapDataToForms` method. See https://github.com/symfony/form/blob/5.2/DataMapperInterface.php and https://github.com/symfony/form/blob/5.3/DataMapperInterface.php
In the configuration of Rector in `config/sets/symfony/symfony50-types.php` it is specified as using `iterable`, but after that there is no update to `\Traversable`.

Could you please add an update for this change in Symfony forms?

johanadivare:
I can confirm this is the case https://github.com/symfony/form/blob/5.2/DataMapperInterface.php has `iterable` @alleknalle you can create pull request for this. Otherwise i can do it

alleknalle:
@johanadivare I've forked this repo and tried to get my fork running locally, but I'm missing a lot of type hints on PHPStan.
Having some issues working that way, but I managed to create a PR, see https://github.com/rectorphp/rector-symfony/pull/358. 
Hope that it's correct this way. I have no clue how to test this as well...

johanadivare:
I see your pull request is merged: https://github.com/rectorphp/rector-symfony/pull/358 good job! You can close this issue.

samsonasik:
Thank you @alleknalle @johanadivare

PR #358, opened 2023-02-13
Change param in Datamapper to Traversable

alleknalle:
See https://github.com/rectorphp/rector-symfony/issues/352

### DIFF
--- a/config/sets/symfony/symfony53.php
+++ b/config/sets/symfony/symfony53.php
@@ -97,12 +97,14 @@ return static function (RectorConfig $rectorConfig): void {
     ]);
 
     $rectorConfig->ruleWithConfiguration(AddParamTypeDeclarationRector::class, [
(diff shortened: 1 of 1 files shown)


config/sets/symfony/symfony53.php
```diff
@@ -91,6 +96,21 @@
...
+            0,
+            new ObjectType(Traversable::class)
+        ),
+        new AddParamTypeDeclaration(
+            'Symfony\Component\Form\DataMapperInterface',
+            'mapDataToForms',
+            1,
+            new ObjectType(Traversable::class)
+        ),
+    ]);
```
TomasVotruba, line 112: Hi, thanks for the fix. Could you add a comment with this feature here?

alleknalle: done!

TomasVotruba: Thanks :+1:

PR #365, opened 2023-02-14
Do not include invalid file in symfony53 set

### LINKED ISSUE #364: Seems to be broken in rector 0.15.5?
The file being included, added in https://github.com/rectorphp/rector-symfony/pull/358/files#diff-f480b78b4fb61aa7087e89ccf55c40fff94c5111b885a08626c3b4382c0b936cR27, does not exist.

marmichalski:
Closes #364.

### DIFF
--- a/config/sets/symfony/symfony53.php
+++ b/config/sets/symfony/symfony53.php
@@ -24,7 +24,7 @@ use Rector\TypeDeclaration\ValueObject\AddReturnTypeDeclaration;
 return static function (RectorConfig $rectorConfig): void {
     $rectorConfig->sets([SymfonySetList::ANNOTATIONS_TO_ATTRIBUTES]);
 
-    $rectorConfig->import(__DIR__ . '/symfony53-types.php');
+    $rectorConfig->import(__DIR__ . '/symfony50-types.php');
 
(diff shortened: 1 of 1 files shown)


config/sets/symfony/symfony53.php
```diff
@@ -24,7 +24,7 @@
 return static function (RectorConfig $rectorConfig): void {
     $rectorConfig->sets([SymfonySetList::ANNOTATIONS_TO_ATTRIBUTES]);
 
-    $rectorConfig->import(__DIR__ . '/symfony53-types.php');
+    $rectorConfig->import(__DIR__ . '/symfony50-types.php');
```
marmichalski, line 27: I am not entirely sure if this should be even included as part of this set?

TomasVotruba: Not at all, it should be removed.

I'm surprised PHPStan missed this.

config/sets/symfony/symfony53.php
```diff
@@ -24,7 +24,7 @@
...
```
marmichalski, line 28:
```suggestion
```

TomasVotruba:
Thanks :+1:

PR #373, opened 2023-02-22
Add tests for overriding constructors in GetToConstructorInjectionRector

jackbentley:


samsonasik:
I am closing it as it can be resolved by using rule among with AddConstructorParentCallRector

- https://github.com/rectorphp/rector-src/pull/3430

jackbentley:
@samsonasik Tests still fail when adding that rule.

samsonasik:
Ok, reopen then, feel free to provide a patch as well when you have a chance as well. I think the child parent param fill should be configurable per we discussed as `__construct` is magic and the usage can be different on purpose.

samsonasik:
I am closing it, since as previoulsly stated, construct is special, and can be different with parent on purpose, that's the "real" override :) 

Feel free to open new PR if you have proposed safe solution.

PR #375, opened 2023-03-06
Update Symfony62Test to cover to not remove existing use statement on rename class

samsonasik:
This is to cover make consistent to not remove unused use when no auto import (replace) and no `$rectorConfig->removeUnusedImports()` used, see no namespace vs namespaced class below:

**Without namespace** (got removed)  : https://getrector.com/demo/1fe1232a-4741-4564-b225-8f0822338071
**With namespace** (unused use is kept)  : https://getrector.com/demo/b4de0818-8ca3-496b-92c7-6f54e9ffd4cd

on PR:

- https://github.com/rectorphp/rector-src/pull/3460


samsonasik:

TomasVotruba:
Thanks :ok:

PR #378, opened 2023-03-06
[MessageHandlerInterfaceToAttributeRector] Adding check for method

### LINKED ISSUE #376: False positive for `MessageHandlerInterfaceToAttributeRector` when attribute is used on class method
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | 0.15.20             |

**Important:** This only occurs when `rectorConfig->symfonyContainerXml()` and `$rectorConfig->symfonyContainerPhp()` are used and the Symfony container is compiled before running rector.

Which is why I can not reproduce it with https://getrector.org/demo/. ~~I hope the minimal code below is enough for you, otherwise I could put it in a reproducer repo.~~ I added a reproducer repo, see comment below.

```
// src/Message.php
<?php

declare(strict_types=1);

namespace App;

final readonly class Message
{
}
```

```
// src/Test.php
<?php

declare(strict_types=1);

namespace App;

use Symfony\Component\Messenger\Attribute\AsMessageHandler;

final readonly class Test
{
    #[AsMessageHandler]
    public function test(Message $message): void
    {
        // Todo: implement
    }
}
```

```
// rector.php
    $rectorConfig->paths([
            __DIR__ . '/src',
    ]);

    $rectorConfig->symfonyContainerXml('var/cache/dev/App_KernelDevDebugContainer.xml');

    $rectorConfig->symfonyContainerPhp('var/cache/dev/App_KernelDevDebugContainer.php');

    $rectorConfig->rule(MessageHandlerInterfaceToAttributeRector::class);
};
```

Output
```
vendor/bin/rector process --dry-run
 1/1 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%
1 file with changes
===================

1) src/Test.php:5

    ---------- begin diff ----------
@@ @@

 use Symfony\Component\Messenger\Attribute\AsMessageHandler;

+#[AsMessageHandler(bus: '', handles: '', method: 'test', priority: 0, from_transport: '')]
 final readonly class Test
 {
     #[AsMessageHandler]
    ----------- end diff -----------

Applied rules:
 * MessageHandlerInterfaceToAttributeRector

 [OK] 1 file would have changed (dry-run) by Rector
```

## Expected Behaviour

Rector should not do anything here, as the `AsMessageHandler` attribute is allowed on methods since Symfony 6.1 (https://symfony.com/doc/current/messenger.html#creating-a-message-handler).

stefantalen:
Fixes #376

### DIFF
--- a/docs/rector_rules_overview.md
+++ b/docs/rector_rules_overview.md
@@ -1,4 +1,4 @@
-# 79 Rules Overview
+# 80 Rules Overview
 
 ## ActionSuffixRemoverRector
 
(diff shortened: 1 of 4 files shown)


TomasVotruba:
Thanks 🙏

gndk:
@stefantalen Thanks for the quick fix! 🚀

PR #379, opened 2023-03-14
Drop triun/longest-common-substring dependency

staabm:
closes https://github.com/rectorphp/rector/issues/7832

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -37,6 +37,9 @@ parameters:
              message: '#Instead of "instanceof/is_a\(\)" use ReflectionProvider service or "\(new ObjectType\(<desired_type\>\)\)\-\>isSuperTypeOf\(<element_type\>\)" for static reflection to work#'
              path: src/ValueObjectFactory/ServiceMapFactory.php
 
+        - '#Cognitive complexity for "Rector\\Symfony\\Rector\\Closure\\MinimalSharedStringSolver\:\:solve\(\)" is (.*?), keep it under 10#'
+
+
         # closure/callable
         - '#Parameter \#2 \$filter of method Rector\\Core\\PhpParser\\Node\\BetterNodeFinder\:\:findFirstPrevious\(\) expects callable\(PhpParser\\Node\)\: bool, Closure\(PhpParser\\Node\)\: PhpParser\\Node\\Expr\\Assign\|null given#'
 
--- a/src/Rector/Closure/MinimalSharedStringSolver.php
+++ b/src/Rector/Closure/MinimalSharedStringSolver.php
@@ -9,10 +9,7 @@ namespace Rector\Symfony\Rector\Closure;
  */
 final class MinimalSharedStringSolver
 {
-    /**
-     * @return string
-     */
-    public function solve(string $stringA, string $stringB)
+    public function solve(string $stringA, string $stringB): string
     {
         if (func_num_args() > 2) {
             $arguments = func_get_args();
@@ -55,10 +52,8 @@ final class MinimalSharedStringSolver
 
     /**
      * @param array<int, array<int, int>> $matrix
-     *
-     * @return int
      */
-    private function newIndex(array $matrix, int $i, int $j)
+    private function newIndex(array $matrix, int $i, int $j): int
     {
         return $i - $matrix[$i][$j] + 1;
     }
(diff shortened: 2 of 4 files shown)


staabm:
phpstan reports

>  -1     Out of 882 possible return types, only 99 % actually have it. Add more return types to get over 100 %

but I don't see a single method which does not declare its return type?

phpstan.neon
```diff
@@ -37,6 +37,9 @@ parameters:
              message: '#Instead of "instanceof/is_a\(\)" use ReflectionProvider service or "\(new ObjectType\(<desired_type\>\)\)\-\>isSuperTypeOf\(<element_type\>\)" for static reflection to work#'
              path: src/ValueObjectFactory/ServiceMapFactory.php
 
+        - '#Cognitive complexity for "Rector\\Symfony\\Rector\\Closure\\MinimalSharedStringSolver\:\:solve\(\)" is (.*?), keep it under 10#'
```
staabm, line 40: I decided to ignore this error, since I don't want to fiddle in this "external" code and introduce bugs since we don't know all the details

src/Rector/Closure/MinimalSharedStringSolver.php
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Class Solver copied from https://github.com/Triun/PHP-Longest-Common-Substring
+ */
+final class MinimalSharedStringSolver
+{
+    /**
+     * @return string
+     */
+    public function solve(string $stringA, string $stringB)
```
samsonasik, line 15: add return type string here

src/Rector/Closure/MinimalSharedStringSolver.php
```diff
@@ -0,0 +1,75 @@
...
+
+        return $this->result($longestIndexes, $longestLength, $stringA);
+    }
+
+    /**
+     * @param array<int, array<int, int>> $matrix
+     *
+     * @return int
+     */
+    private function newIndex(array $matrix, int $i, int $j)
```
samsonasik, line 61: add return type int here

src/Rector/Closure/MinimalSharedStringSolver.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        return $i - $matrix[$i][$j] + 1;
+    }
+
+    /**
+     * @param list<int> $longestIndexes
+     *
+     * @return string the extracted part of string or false on failure.
+     */
+    private function result(array $longestIndexes, int $longestLength, string $stringA)
```
samsonasik, line 71: add return type string here

src/Rector/Closure/ServiceSettersToSettersAutodiscoveryRector.php
```diff
@@ -106,9 +105,6 @@ public function refactor(Node $node): ?Node
         );
 
         $sharedNamespace = $this->minimalSharedStringSolver->solve(...$classNames);
-        if (! is_string($sharedNamespace)) {
-            return null;
-        }
```
staabm, line 111: since the `MinimalSharedStringSolver` no longer implements the `Solver` interface, we were able to declare more specific return types which also means this case - which was never taken, as `null` could technically not be returned - can be dropped.

src/Rector/Closure/MinimalSharedStringSolver.php
staabm: ahh didn't realize that I need to use native return types. maybe the error message should be more precise.

thanks - fixed.

samsonasik (approved):

samsonasik:
Thank you @staabm

src/Rector/Closure/MinimalSharedStringSolver.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Symfony\Rector\Closure;
+
+/**
+ * Class Solver copied from https://github.com/Triun/PHP-Longest-Common-Substring
+ */
+final class MinimalSharedStringSolver
```
TomasVotruba, line 10: @staabm This needs a test, as there is lot of tiny coding happening here.
Could you add it?

staabm: sure, here you are: https://github.com/rectorphp/rector-symfony/pull/380

PR #382, opened 2023-03-22
Performance: Reduce isObjectType calls

keulinho:
Apply changes from https://github.com/rectorphp/rector-src/pull/3502 here as well


samsonasik (approved):
Looks good, thank you @keulinho  👍

PR #394, opened 2023-04-26
Skip RemoveUnusedRequestParamRector on override abstract method

### LINKED ISSUE #366: `RemoveUnusedRequestParamRector` causes bad PHP code when using abstracts
# Bug Report

See https://getrector.com/demo/381c0067-b8fa-4c7b-8bb1-80f0737a4799

```php
<?php

namespace Symfony\Bundle\FrameworkBundle\Controller {
    class Controller {}
}
    
namespace Symfony\Component\HttpFoundation {
    class Request {}
    class Response {}
}

namespace App\Controller {
    use Symfony\Bundle\FrameworkBundle\Controller\Controller;
    use Symfony\Component\HttpFoundation\Request;
    use Symfony\Component\HttpFoundation\Response;

    abstract class AbstractController extends Controller
    {
        abstract public function fooAction(Request $request): Response;
    }

    class DemoController extends AbstractController
    {
        public function fooAction(Request $request): Response
        {
            return new Response();
        }

        public function barAction(Request $request): Response
        {
            return new Response();
        }
    }
}
```

### Responsible rules

* `RemoveUnusedRequestParamRector`

## Expected Behavior

If the method is an override then it's parameters should not be altered. This would result in inoperable PHP code.

The `$request` parameter should remain in the `fooAction` otherwise the PHP code is inoperable.

JohJohan:
Fixes #366

### DIFF
--- a/src/Rector/ClassMethod/RemoveUnusedRequestParamRector.php
+++ b/src/Rector/ClassMethod/RemoveUnusedRequestParamRector.php
@@ -7,8 +7,11 @@ namespace Rector\Symfony\Rector\ClassMethod;
 use PhpParser\Node;
 use PhpParser\Node\Expr\Variable;
 use PhpParser\Node\Stmt\ClassMethod;
+use PHPStan\Reflection\ClassReflection;
 use PHPStan\Type\ObjectType;
 use Rector\Core\Rector\AbstractRector;
+use Rector\Core\Reflection\ReflectionResolver;
+use Rector\FamilyTree\NodeAnalyzer\ClassChildAnalyzer;
 use Rector\Symfony\TypeAnalyzer\ControllerAnalyzer;
 use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
 use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
@@ -19,7 +22,9 @@ use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
 final class RemoveUnusedRequestParamRector extends AbstractRector
 {
     public function __construct(
-        private readonly ControllerAnalyzer $controllerAnalyzer
+        private readonly ControllerAnalyzer $controllerAnalyzer,
+        private readonly ReflectionResolver $reflectionResolver,
+        private readonly ClassChildAnalyzer $classChildAnalyzer,
     ) {
     }
 
@@ -74,6 +79,10 @@ CODE_SAMPLE
             return null;
         }
 
+        if ($node->isAbstract() || $this->hasAbstractParentClassMethod($node)) {
+            return null;
+        }
+
         if (! $this->controllerAnalyzer->isInsideController($node)) {
             return null;
         }
@@ -112,5 +121,15 @@ CODE_SAMPLE
         }
 
         return null;
+    }
+
+    private function hasAbstractParentClassMethod(ClassMethod $classMethod): bool
+    {
+        $classReflection = $this->reflectionResolver->resolveClassReflection($classMethod);
+        if (! $classReflection instanceof ClassReflection) {
+            return false;
+        }
+
+        return $this->classChildAnalyzer->hasAbstractParentClassMethod($classReflection, $this->getName($classMethod));
     }
 }
--- a/src/NodeAnalyzer/FormAddMethodCallAnalyzer.php
+++ b/src/NodeAnalyzer/FormAddMethodCallAnalyzer.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Rector\Symfony\NodeAnalyzer;
 
+use PhpParser\Node\Expr;
 use PhpParser\Node\Expr\MethodCall;
 use PHPStan\Type\ObjectType;
 use Rector\NodeNameResolver\NodeNameResolver;
@@ -43,6 +44,6 @@ final class FormAddMethodCallAnalyzer
         }
 
         $firstArg = $args[1];
-        return $firstArg->value !== null;
+        return $firstArg->value instanceof Expr;
     }
 }
--- a/src/NodeAnalyzer/LiteralCallLikeConstFetchReplacer.php
+++ b/src/NodeAnalyzer/LiteralCallLikeConstFetchReplacer.php
@@ -49,9 +49,7 @@ final class LiteralCallLikeConstFetchReplacer
             return null;
         }
 
-        $classConstFetch = $this->nodeFactory->createClassConstFetch($className, $constantName);
-
-        $arg->value = $classConstFetch;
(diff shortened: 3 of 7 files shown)


johanadivare:
Fixes #366

johanadivare:
I dont know how to fix the rector failing, that has probably nothing to do with my changes.

The PHPstan fails because undefined method but it does work can i add it to the ignores?

TomasVotruba:
@samsonasik It seems the BetterTokenIterator::getTokens() was removed. Do you have any idea how this happened?

samsonasik:
Latest php-doc-parser already have it, so I removed it as it extends the `getTokens()` in rector-src as extends it, see 

- https://github.com/rectorphp/rector-src/pull/3655
- https://github.com/phpstan/phpdoc-parser/commit/900bd69a93e9bd51a11a0bdc6915c4bd813b25c3#diff-62d19c66cc22434f96724388acd962198a200c0c0ced48aeace88157ae658586

php-doc-parser may need to be updated to at least version `^1.20.1`

TomasVotruba:
Ah, I see now 😁

samsonasik:
Let's restart build.

samsonasik:
I will look into why it not load latest phpdoc-parser, @JohJohan  I may cherry-pick your commit in case I can't push to your branch :)

JohJohan:
> I will look into why it not load latest phpdoc-parser, @JohJohan  I may cherry-pick your commit in case I can't push to your branch :) 

Sounds good, good luck :)

samsonasik:
It seems the autoload is overlapped with phpstan.phar's inside build release 1.10.4 which still uses php-doc-parser pinned to `1.19.1`, see https://github.com/phpstan/phpstan-src/blob/532094bc1b70b462c093bfc17eb47b149ff6cb0e/composer.json#L28

rector-src provide preload.php and preload-split-package.php, that's why it works there, I will check if it can be tweaked here :)

samsonasik:
see https://github.com/rectorphp/rector-src/pull/3692

samsonasik:
Let's restart build ...

samsonasik:
@JohJohan I fixed the rector error :tada: , I will let you fix the phpstan notice part :)

phpstan.neon
```diff
@@ -50,6 +50,10 @@ parameters:
         # known type
         - '#Parameter \#1 \$className of method Rector\\Core\\Reflection\\ReflectionResolver\:\:resolveMethodReflection\(\) expects class\-string, string given#'
 
+        -
+            message: '#Call to an undefined method PHPStan\\Reflection\\ExtendedMethodReflection\:\:isAbstract\(\)#'
```
samsonasik, line 54: This should be fixed with set actual object with `isAbstract()` method, you can debug actual object with `dump($variable)`

johanadivare: The actual class is `PHPStan\Reflection\Php\PhpMethodReflection`  which has `isAbstract` but when i do something like `assert($extendedMethodReflection instanceof PHPStan\Reflection\Php\PhpMethodReflection);` i get error ` Class Rector\Symfony\Rector\ClassMethod\PHPStan\Reflection\Php\PhpMethodReflection not found. `

samsonasik: You need to use FQCN, or that will refer current namespace:

```diff
-assert($extendedMethodReflection instanceof PHPStan\Reflection\Php\PhpMethodReflection); 
+assert($extendedMethodReflection instanceof \PHPStan\Reflection\Php\PhpMethodReflection); 
```

johanadivare: Aah thanks :)

samsonasik (changes requested):
Please update PR title to something like:

```
Skip RemoveUnusedRequestParamRector on override abstract method
```

src/Rector/ClassMethod/RemoveUnusedRequestParamRector.php
```diff
@@ -113,4 +124,34 @@ public function refactor(Node $node): ?Node
...
+            return false;
+        }
+        $reflectionParentClass = $classReflection->getParentClass();
+        if (! $reflectionParentClass instanceof ClassReflection) {
+            return false;
+        }
+        try {
+            $extendedMethodReflection = $reflectionParentClass->getNativeMethod($methodName);
+            assert($extendedMethodReflection instanceof PhpMethodReflection);
+            if (! $extendedMethodReflection->isAbstract()) {
```
samsonasik, line 148: There is already service to verify it, see

https://github.com/rectorphp/rector-src/blob/5c1be93b80c3f7bfd8bd158ba0a55501a7acaefc/packages/FamilyTree/NodeAnalyzer/ClassChildAnalyzer.php#L52

johanadivare: Nice find, i thought there must have been something but i couldn't find it i keep my private function to do the reflection and checking see `hasAbstractParentClassMethod` is that okay?

samsonasik: That's fine as you need to get `ClassReflection` instance 👍

tests/Rector/ClassMethod/RemoveUnusedRequestParamRector/Fixture/controller.php.inc
```diff
@@ -0,0 +1,55 @@
...
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+abstract class AbstractController extends Controller
+{
+    abstract public function fooAction(Request $request): Response;
+}
+
+class DemoController extends AbstractController
```
samsonasik, line 14: It needs to be more specific name so it ease to look on the future, eg:

```suggestion
class DontOverrideParentClassMethodController extends AbstractController
```

tests/Rector/ClassMethod/RemoveUnusedRequestParamRector/Fixture/controller.php.inc
```diff
@@ -0,0 +1,55 @@
+<?php
```
samsonasik, line 1: please rename fixture file to something like: `do_not_override_parent_abstract_class_method.php.inc`

johanadivare: ++

tests/Rector/ClassMethod/RemoveUnusedRequestParamRector/Fixture/controller.php.inc
johanadivare: ++

samsonasik (approved):

samsonasik:
Thank you @JohJohan

Issue #184, opened 2022-06-13
FormTypeWithDependencyToOptionsRector.php is ruining form types that are intended to be services


trsteel88:
If we're injecting a service into a form type, rector is attempting to rewrite the parameter as an option.

This shouldn't happen because we need a service injected from the container. Is there a reason this is a default rule?

TomasVotruba:
Hi, could you share exact link with your use case?
https://getrector.org/demo

Are you running Symfony 2.8 upgrade on latter versions?

trsteel88:
@TomasVotruba sorry, I'm not sure how to do that in there but here is a codebase example which is having an issue:

https://github.com/trsteel88/rector-test

```
$ ./vendor/bin/rector process
 3/3 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%
                                                                                                                        
 [ERROR] Could not process "src/Form/Choice/ApiChoice.php" file, due to:                                                
         "System error: "Look at "Rector\Symfony\Rector\Class_\FormTypeWithDependencyToOptionsRector::refactor()":115"  
         Run Rector with "--debug" option and post the report here: https://github.com/rectorphp/rector/issues/new". On 
         line: 115     
```

PR #467, opened 2023-06-24
Move FormTypeWithDependencyToOptionsRector to Symfony 3.0 set

TomasVotruba:


W0rma:
@TomasVotruba What's the motivation behind this change?

Since https://github.com/symfony/symfony-docs/commit/ac08120d4985b9647e319f5b10b1cb2cb248391e#diff-3f7adba991201416ba5b300423ce923540a4374c6b5e7fa3ff128a2055bf28b0 even the symfony docs recommend using dependency injection.

TomasVotruba:
I see. Here is it depends whether the object is a dependency service or value object.

Could you add a test fixture that fails for service?

W0rma:
I have added a test case in https://github.com/rectorphp/rector-symfony/pull/488

bobvandevijver:
This was already discussed before in #180, #184 and fixed with #189. The FormTypeWithDependencyToOptionsRector should be a one-time use for people who want to use it, so I am quite surprised to see it has now been readded to the Symfony set.

TomasVotruba:
@bobvandevijver Thanks for joining in :+1: my bad and appologies for caused trouble.s

I wonder if the rule should even exist, if it always require a developers attention and careful consideration of change.
I think it would be better to remove it to avoid similar issues like you've reffered.

PR #489, opened 2023-07-04
Fix SimplifyFormRenderingRector replacing calls other than createView()

adlpz:
The SimplifyFormRenderingRector was replacing calls to methods other than ->createView(), which is no longer required in Symfony 6.2, e.g. calls to ->getData().

These are not deprecated and there's no reason to remove them, and in fact the blanket rule of removing any method calls of FormInterface parameters of the container $this->render() function is too wide and could lead to breakage.


samsonasik (changes requested):
please add test case at https://github.com/rectorphp/rector-symfony/tree/main/rules-tests/Symfony62/Rector/MethodCall/SimplifyFormRenderingRector/Fixture

adlpz:
Wooops, I uploaded the base test case I wrote first. Fixed.

samsonasik (approved):

adlpz:
Wooops, I uploaded the base test case I wrote first. Fixed.

PR #490, opened 2023-07-06
[Symfony 3.0] Remove FormTypeWithDependencyToOptionsRector, as it always require manual check of code

TomasVotruba:
Ref issue caused by this rule https://github.com/rectorphp/rector-symfony/pull/467#issuecomment-1619457378

Instead, it should be handled manually on per-case basis :+1:


bobvandevijver:
I think this is a great move. 👍🏻 This rector isn't even relevant anymore for newer Symfony versions, as the form type must be passed as class-string, and never as instance.

TomasVotruba:
@bobvandevijver Perfect, thanks for constructive feedback on this one :+1:

PR #505, opened 2023-08-10
[DX] Remove PseudoNamespaceToNamespace config in twig set, use explicit class rename directly as better and reliable rule

TomasVotruba:


config/sets/twig/twig-underscore-to-namespace.php
```diff
@@ -163,14 +162,35 @@
...
+        'Twig_Lexer' => 'Twig\\Lexer',
+        'Twig_Markup' => 'Twig\\Markup',
+        'Twig_Parser' => 'Twig\\Parser',
+        'Twig_Source' => 'Twig\\Source',
+        'Twig_Template' => 'Twig\\Template',
+        'Twig_Token' => 'Twig\\Token',
+        'Twig_TokenStream' => 'Twig\\TokenStream',
+        'Twig_Extension_GlobalsInterface' => 'Twig\\Extension\\GlobalsInterface',
+        'Twig_Extension_InitRuntimeInterface' => 'Twig\\Extension\\InitRuntimeInterface',
+        'Twig_Node_Deprecated' => 'Twig\\Node\\DeprecatedNode',
```
TomasVotruba, line 178: Appearantly, there were few bugs in node renames, like this one :)

That's why explicit list is always better to make sure we have the values right :+1:

PR #511, opened 2023-08-20
Remove symfony/strings dependency

TomasVotruba:


rules/Symfony30/Rector/ClassMethod/RemoveDefaultGetBlockPrefixRector.php
```diff
@@ -118,6 +114,14 @@ public function refactor(Node $node)
         return null;
     }
 
+    private function camelToSnake(string $content): string
+    {
+        return strtolower(
```
samsonasik, line 119: @TomasVotruba class name allow contains unicode, so it needs to use `mb_strtolowe()`, see https://3v4l.org/jQJGM

samsonasik: https://github.com/rectorphp/rector-symfony/pull/512

PR #513, opened 2023-08-22
Fix test on symfony-44 test on merge of symfony 4.3 + symfony 4.4 ruleWithConfiguration()

samsonasik:
@TomasVotruba @lucasmirloup this is based on fix of PR:

- https://github.com/rectorphp/rector-src/pull/4831

temporary use `rector-src:dev-failing-fixture` for it.


samsonasik:
Oh, it's not, `symfony-44` is not merge of `symfony-43` and `symfony-44` so when enter new test class, it should reset the `Option::RECTOR_CONFIGURATION` values. Closing for now.

PR #519, opened 2023-09-07
[Symfony43] Skip not instanceof FilesystemLoader via dynamic new $class on TwigBundleFilesystemLoaderToTwigRector

samsonasik:
Fixes https://github.com/rectorphp/rector/issues/8189

### DIFF
--- a/rules/Symfony43/Rector/StmtsAwareInterface/TwigBundleFilesystemLoaderToTwigRector.php
+++ b/rules/Symfony43/Rector/StmtsAwareInterface/TwigBundleFilesystemLoaderToTwigRector.php
@@ -6,6 +6,7 @@ namespace Rector\Symfony\Symfony43\Rector\StmtsAwareInterface;
 
 use PhpParser\Node;
 use PhpParser\Node\Arg;
+use PhpParser\Node\Expr;
 use PhpParser\Node\Expr\Assign;
 use PhpParser\Node\Expr\MethodCall;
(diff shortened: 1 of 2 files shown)


rules/Symfony43/Rector/StmtsAwareInterface/TwigBundleFilesystemLoaderToTwigRector.php
```diff
@@ -107,6 +108,10 @@ private function resolveFileSystemLoaderNew(StmtsAwareInterface $stmtsAware): ?N
                 continue;
             }
 
+            if ($assign->expr->class instanceof Expr) {
+                continue;
+            }
```
samsonasik, line 113: @TomasVotruba the `new $class` got true on `isObjectType()` so this check is needed here, the `isObjectType()` method itself may need improvement to detect it automatically.

samsonasik:
Fixed 🎉 /cc @Knallcharge

samsonasik:

rules/Symfony43/Rector/StmtsAwareInterface/TwigBundleFilesystemLoaderToTwigRector.php
samsonasik: @TomasVotruba I created PR to allow cover on `NewTypeResolver` on PR:

- https://github.com/rectorphp/rector-src/pull/4936

PR #523, opened 2023-09-09
ResponseStatusCodeRector: use NodeNameResolver->matchesStringName()

staabm:
counterpart of https://github.com/rectorphp/rector-src/pull/4951


TomasVotruba:
Thanks :+1:

PR #527, opened 2023-09-21
Adding handling of method chaining for MakeCommandLazyRector

### LINKED ISSUE #526: MakeCommandLazy rule is broken
I have a configure method like this:

```
   protected function configure(): void
    {
        $this
            ->setName('some-command')
            ->addOption(
                'some_option',
                null,
                InputOption::VALUE_OPTIONAL,
                'Some text'
            )
            ->setDescription('some_description')
            ->addOption('some_other_option', 'i', InputOption::VALUE_NONE, 'Some other text');
    }
```

Running the `MakeCommandLazy` rule doesn't do anything here. But I found out that if I split the definition like this:

```
   protected function configure(): void
    {
        $this->setName('some-command');
        $this->addOption(
                'some_option',
                null,
                InputOption::VALUE_OPTIONAL,
                'Some text'
            )
            ->setDescription('some_description')
            ->addOption('some_other_option', 'i', InputOption::VALUE_NONE, 'Some other text');
    }
```

Then the rule works as expected. So apparently the rule cannot be applied when method chaining is used. Similarly there another bug, if `setName` is the last method in the chain:

```
   protected function configure(): void
    {
        $this
            ->addOption(
                'some_option',
                null,
                InputOption::VALUE_OPTIONAL,
                'Some text'
            )
            ->setDescription('some_description')
            ->addOption('some_other_option', 'i', InputOption::VALUE_NONE, 'Some other text')
            ->setName('some-command');
    }
```

While this does work in creating the property `defaultName` from the `setName` function, it wrongfully deletes all other configurations, leaving me with an empty `configure` method.

stefantalen:
fixes #526

### DIFF
--- a/rules-tests/CodeQuality/Rector/Class_/MakeCommandLazyRector/Fixture/method_chaining.php.inc
+++ b/rules-tests/CodeQuality/Rector/Class_/MakeCommandLazyRector/Fixture/method_chaining.php.inc
@@ -1,0 +1,41 @@
+<?php
+
+namespace Rector\Symfony\Tests\CodeQuality\Rector\Class_\MakeCommandLazyRector\Fixture;
+
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class MethodChainingCommand extends \Symfony\Component\Console\Command\Command
+{
+    public function configure()
+    {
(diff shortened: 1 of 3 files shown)


TomasVotruba:
Awesome, this looks great. Thanks :+1:

PR #541, opened 2023-10-25
Fix AssertSameResponseCodeWithDebugContentsRector to use getContent instead of getContents

alexander-schranz:
Not sure how this was even tested but the `getContents` method does not exist on the [Symfony HttpFoundation Response](https://github.com/symfony/symfony/blob/b2282614e5316e55a096124f8e82b742eccfb79d/src/Symfony/Component/HttpFoundation/Response.php#L459):

PHPStan erros with:

> Call to an undefined method Symfony\Component\HttpFoundation\Response::getContents().

Technically the rule should be also renamed from `AssertSameResponseCodeWithDebugContentsRector` -> `AssertSameResponseCodeWithDebugContentRector` but that would be a bc break.

### DIFF
--- a/docs/rector_rules_overview.md
+++ b/docs/rector_rules_overview.md
@@ -101,7 +101,7 @@ Make assertSame(200, `$response->getStatusCode())` in tests comparing response c
          $response = $this->processResult();
 
 -        $this->assertSame(200, $response->getStatusCode());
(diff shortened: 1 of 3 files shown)


TomasVotruba:
> or maybe cast:

That would be best IMO :+1:

PR #542, opened 2023-10-25
Temporary use rector-src:dev-rever-doctrine-inflector

samsonasik:


samsonasik:
Don't work, I need your hand to manually merge https://github.com/rectorphp/rector-src/pull/5201

Issue #535, opened 2023-10-12
MessageHandlerInterfaceToAttributeRector False positive


jambonfarci:
Rector demo link : [https://getrector.com/demo/1e2e5f56-7ca6-40fb-94bf-a1d1cfa13059](https://getrector.com/demo/1e2e5f56-7ca6-40fb-94bf-a1d1cfa13059)

The changes don't match with the rule applied : 
![image](https://github.com/rectorphp/rector-symfony/assets/1522511/c296bcfb-6b74-4b9e-a32d-54e9da2954f8)

TomasVotruba:
Thank you for your report and demo link!

We'll need a failing test case in a pull-request, so we have it covered in Rector.
You can click "Create a Test" button at demo page.

jambonfarci:
Thanks for your quick reply !
I'll try to covered the test as soon as possible.

jambonfarci:
I've narrowed down the rules to be more precise : [https://getrector.com/demo/19b6ba38-d974-4095-a5f2-576cad86171f](https://getrector.com/demo/19b6ba38-d974-4095-a5f2-576cad86171f)

I'll try to make a PR with a failing test next.

stefantalen:
Looks like an issue with combined rules.

Just using the `MessageHandlerInterfaceToAttributeRector` results in no changes 🙂 

Can you narrow down the rules instead of using a set?

jambonfarci:
Indeed, i've found the rules  : [https://getrector.com/demo/e761b19d-3501-4f1d-85b1-51a9d1fca87c](https://getrector.com/demo/e761b19d-3501-4f1d-85b1-51a9d1fca87c)

The `RenameClassRector` rule is part of the Symfony6.2 Set.

stefantalen:
@TomasVotruba should this be considered a `rector/rector` issue?

samsonasik:
@stefantalen it seems yes, I will look into it

samsonasik:
Looking at how transformation of `IsGranted`, it  tries to rename:

https://github.com/rectorphp/rector-symfony/blob/873458ca73a73f9becd1680963a347bbcb8e5f43/config/sets/symfony/symfony62.php#L24-L25

while it **still** an `Annotation`. On other hand, the `sensiolabs` is changing to attribute on `Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted` instead (old class).

[...]

PR #544, opened 2023-10-29
[Symfony 62] Ensure change Annotation to Attribute first before rename Annotation on IsGranted, Cache, Template

samsonasik:
@jambonfarci @stefantalen @TomasVotruba here ensure change annotation to attribute before rename,

see https://getrector.com/demo/894cda53-f2a9-4216-b64a-7ba2247b0a73
Ref https://github.com/rectorphp/rector-symfony/issues/535#issuecomment-1783983383

For auto import, it require PR on rector-src on auto import enabled usage:

- https://github.com/rectorphp/rector-src/pull/5213


samsonasik:

PR #545, opened 2023-10-31
Fix https://github.com/rectorphp/rector-symfony/issues/352

alleknalle:
Looks like https://github.com/rectorphp/rector-symfony/issues/352 was reverted somewhere...

### DIFF
--- a/config/sets/symfony/symfony50-types.php
+++ b/config/sets/symfony/symfony50-types.php
@@ -295,14 +295,14 @@ return static function (RectorConfig $rectorConfig): void {
             'Symfony\Component\Form\DataMapperInterface',
             'mapFormsToData',
             0,
-            new ObjectType(Traversable::class)
+            new ObjectType('Traversable')
         ),
         // @see https://github.com/symfony/symfony/commit/ce77be2507631cd12e4ca37510dab37f4c2b759a
         new AddParamTypeDeclaration(
             'Symfony\Component\Form\DataMapperInterface',
             'mapDataToForms',
             1,
-            new ObjectType(Traversable::class)
+            new ObjectType('Traversable')
         ),
         new AddParamTypeDeclaration('Symfony\Component\Form\Form', 'add', 1, new StringType()),
         new AddParamTypeDeclaration('Symfony\Component\Form\Form', 'remove', 0, new StringType()),
--- a/config/sets/symfony/symfony50-types.php
+++ b/config/sets/symfony/symfony50-types.php
@@ -289,17 +289,19 @@ return static function (RectorConfig $rectorConfig): void {
             0,
             new StringType()
         ),
+        // @see https://github.com/symfony/symfony/commit/ce77be2507631cd12e4ca37510dab37f4c2b759a
         new AddParamTypeDeclaration(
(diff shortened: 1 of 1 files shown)


config/sets/symfony/symfony50-types.php
```diff
@@ -289,17 +289,19 @@
             0,
             new StringType()
         ),
+        // @see https://github.com/symfony/symfony/commit/ce77be2507631cd12e4ca37510dab37f4c2b759a
         new AddParamTypeDeclaration(
             'Symfony\Component\Form\DataMapperInterface',
             'mapFormsToData',
             0,
-            $iterableType
+            new ObjectType(Traversable::class)
```
samsonasik, line 297: use string `"Traversable"` instead of `::class` to avoid prefixed on downgrade

config/sets/symfony/symfony50-types.php
```diff
@@ -289,17 +289,19 @@
...
         ),
+        // @see https://github.com/symfony/symfony/commit/ce77be2507631cd12e4ca37510dab37f4c2b759a
         new AddParamTypeDeclaration(
             'Symfony\Component\Form\DataMapperInterface',
             'mapDataToForms',
             1,
-            $iterableType
+            new ObjectType(Traversable::class)
```
samsonasik, line 304: use string `"Traversable"` instead of `::class` to avoid prefixed on downgrade

samsonasik:
I am on mobile, you may need to check git blame to check why it was changed

alleknalle:
@samsonasik It looks like when I use the specific ``SymfonySetList::SYMFONY_53`` it works, but when I use one of the ``UP_TO_...`` SymfonyLevelSetLists it doesn't. 

Can it be an order problem?

In ``symfony50-types.php`` it states that the type of the first parameter in ``Symfony\Component\Form\DataMapperInterface::mapFormsToData()`` needs to be of ``iterable`` type, but in ``symfony53.php`` it is updated to ``Traversable``. 

Locally I've changed the ``UP_TO_SYMFONY_53`` list to NOT use the ``UP_TO_SYMFONY_52`` list and then it works correctly. 
I've also tried it with defining ``UP_TO_SYMFONY_52`` BEFORE ``SYMFONY_53`` in ``up-to-symfony-53.php`` and then it works as well.

So to me it looks like it is some order issue.

samsonasik:
Reproducible test case seems needed for that.

alleknalle:
@samsonasik I've created a scenario in a new repo. Please see https://github.com/alleknalle/rector-test.
Is there anything else I can do to check this?

samsonasik:
Could you create failing test case in this repo under `tests/` https://github.com/rectorphp/rector-symfony/tree/main/tests/Set ?

That will be easier to see what the expected output instead of reproduce in the whole symfony project.

alleknalle:
> Could you create failing test case in this repo under `tests/` https://github.com/rectorphp/rector-symfony/tree/main/tests/Set ?
> 
> That will be easier to see what the expected output instead of reproduce in the whole symfony project.

@samsonasik done, see https://github.com/rectorphp/rector-symfony/pull/547

Issue #401, opened 2023-05-24
SwiftSetBodyToHtmlPlainMethodCallRector example makes use of ->text but inside code it uses ->plain which afaik doesnt exist


torian257x:
https://github.com/rectorphp/rector-symfony/blob/3fa995c7677b2ea654e17dc96cfed1a9ed83732b/src/Rector/MethodCall/SwiftSetBodyToHtmlPlainMethodCallRector.php#L86

->plain doesnt exist see https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/Mime/Email.php

->text does

other things seem to be wrong too like still handing second argument.

TomasVotruba:
Hi, thanks for noticing. Could you send a fix to this rule?

stefantalen:
Looks like the initial plan was to replace a `Swift_Message` with an `Email`, but things got misaligned.

The current rule cannot be used for such a replacement because there are more stones to be moved.
```php
<?php

$message = (new Swift_Message())
    ->setSubject('Your subject')
    ->setFrom(['john@doe.com' => 'John Doe'])
    ->setTo(['receiver@domain.org', 'other@domain.org' => 'A name'])
    ->setBody('Here is the message itself')
?>
-----
<?php

$message = (new \Symfony\Component\Mime\Email())
    ->subject('Your subject')
    ->from(new \Symfony\Component\Mime\Address('john@doe.com', 'John Doe'))
    ->to(...['receiver@domain.org', new \Symfony\Component\Mime\Address('other@domain.org', 'A name')])
    ->text('Here is the message itself')
?>
```

torian257x:
yeah... no... swift mailer and symfony mailer are just too incompatible for our project.

attachments etc etc we refactored by hand into a service so we wont have to do this ever again.

JohJohan:
@TomasVotruba i think conclusion is to close this issue

TomasVotruba:
@JohJohan Thanks for update :+1:  Closing then

PR #548, opened 2023-12-14
Separating Swiftmailer to Symfony Mailer from Symfony 5.3 set

stefantalen:
I came across these rules while working on a project.

Since these rules are incomplete for their purpose (as mentioned in #401) , I feel like they should not come with a Symfony set.

### DIFF
--- a/config/sets/swiftmailer/swiftmailer-to-symfony-mailer.php
+++ b/config/sets/swiftmailer/swiftmailer-to-symfony-mailer.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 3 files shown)


TomasVotruba:
Makes sense, thank you :+1:

samsonasik:
It seems cause missing in set error on rector-src https://github.com/rectorphp/rector-src/actions/runs/7217050580/job/19664232970?pr=5360#step:5:18

Namespace may need to be renamed or excluded in MissingInSetCommand

samsonasik:
I created new PR to move to SwiftMailer namespace:

- https://github.com/rectorphp/rector-symfony/pull/550

PR #557, opened 2024-01-12
chore(symfony-routing): replace Symfony route annotation by attribute one

mcsky:
Integrate in Symfony upgrade the replacement of Symfony\Component\Routing\Annotation\Route by Symfony\Component\Routing\Attribute\Route

see [this issue](https://github.com/rectorphp/rector/issues/8384)

Closes https://github.com/rectorphp/rector/issues/8384

### DIFF
--- a/config/sets/symfony/symfony64.php
+++ b/config/sets/symfony/symfony64.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 use Rector\Config\RectorConfig;
 use Rector\Renaming\Rector\Name\RenameClassRector;
+use Rector\Symfony\Symfony64\Rector\Class_\ChangeRouteAttributeFromAnnotationSubnamespaceRector;
 
 // @see https://github.com/symfony/symfony/blob/6.4/UPGRADE-6.4.md
 return static function (RectorConfig $rectorConfig): void {
@@ -14,4 +15,5 @@ return static function (RectorConfig $rectorConfig): void {
             'Symfony\Component\HttpKernel\Debug\FileLinkFormatter' => 'Symfony\Component\ErrorHandler\ErrorRenderer\FileLinkFormatter',
         ],
     );
+    $rectorConfig->rule(ChangeRouteAttributeFromAnnotationSubnamespaceRector::class);
 };
--- a/rules-tests/Symfony64/Rector/Class_/ChangeRouteAttributeFromAnnotationSubnamespaceRector/Fixture/use-annotation-should-skip.php.inc
+++ b/rules-tests/Symfony64/Rector/Class_/ChangeRouteAttributeFromAnnotationSubnamespaceRector/Fixture/use-annotation-should-skip.php.inc
@@ -1,0 +1,41 @@
+<?php
+
+namespace Rector\Symfony\Tests\Symfony64\Rector\Class_\ChangeRouteAttributeFromAnnotationSubnamespaceRector\Fixture;
+
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+USE Symfony\Component\Routing\Annotation\Route;
+
+class RectorController
+{
+    /**
+     * @Route("/")
+     */
+    public function __invoke(Request $request): JsonResponse
+    {
+        return new JsonResponse();
+    }
+}
+
+?>
+-----
+<?php
+
+namespace Rector\Symfony\Tests\Symfony64\Rector\Class_\ChangeRouteAttributeFromAnnotationSubnamespaceRector\Fixture;
+
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+USE Symfony\Component\Routing\Annotation\Route;
+
+class RectorController
+{
+    /**
+     * @Route("/")
+     */
+    public function __invoke(Request $request): JsonResponse
+    {
+        return new JsonResponse();
+    }
+}
+
+?>
--- a/rules-tests/Symfony64/Rector/Class_/ChangeRouteAttributeFromAnnotationSubnamespaceRector/Fixture/use-annotation-should-skip.php.inc
+++ b/rules-tests/Symfony64/Rector/Class_/ChangeRouteAttributeFromAnnotationSubnamespaceRector/Fixture/use-annotation-should-skip.php.inc
@@ -16,26 +16,3 @@ class RectorController
         return new JsonResponse();
     }
 }
-
-?>
------
-<?php
-
-namespace Rector\Symfony\Tests\Symfony64\Rector\Class_\ChangeRouteAttributeFromAnnotationSubnamespaceRector\Fixture;
-
-use Symfony\Component\HttpFoundation\JsonResponse;
-use Symfony\Component\HttpFoundation\Request;
-USE Symfony\Component\Routing\Annotation\Route;
-
-class RectorController
-{
-    /**
-     * @Route("/")
-     */
-    public function __invoke(Request $request): JsonResponse
-    {
-        return new JsonResponse();
-    }
-}
-
-?>
--- a/rules-tests/Symfony64/Rector/Class_/ChangeRouteAttributeFromAnnotationSubnamespaceRector/ChangeRouteAttributeFromAnnotationSubnamespaceRectorTest.php
+++ b/rules-tests/Symfony64/Rector/Class_/ChangeRouteAttributeFromAnnotationSubnamespaceRector/ChangeRouteAttributeFromAnnotationSubnamespaceRectorTest.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Symfony\Tests\Symfony64\Rector\Class_\ChangeRouteAttributeFromAnnotationSubnamespaceRector;
+
+use Iterator;
+use PHPUnit\Framework\Attributes\DataProvider;
+use Rector\Testing\PHPUnit\AbstractRectorTestCase;
+
+final class ChangeRouteAttributeFromAnnotationSubnamespaceRectorTest extends AbstractRectorTestCase
+{
+    #[DataProvider('provideData')]
+    public function test(string $filePath): void
+    {
+        $this->doTestFile($filePath);
+    }
+
+    public static function provideData(): Iterator
+    {
+        return self::yieldFilesFromDirectory(__DIR__ . '/Fixture');
+    }
+
+    public function provideConfigFilePath(): string
+    {
+        return __DIR__ . '/config/configured_rule.php';
+    }
+}
--- a/rules-tests/Symfony64/Rector/Class_/ChangeRouteAttributeFromAnnotationSubnamespaceRector/Fixture/use-attribute-with-annotation-import-on-class.php.inc
+++ b/rules-tests/Symfony64/Rector/Class_/ChangeRouteAttributeFromAnnotationSubnamespaceRector/Fixture/use-attribute-with-annotation-import-on-class.php.inc
@@ -1,0 +1,35 @@
+<?php
+
+namespace Rector\Symfony\Tests\Symfony64\Rector\Class_\ChangeRouteAttributeFromAnnotationSubnamespaceRector\Fixture;
+
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+
+#[\Symfony\Component\Routing\Annotation\Route('/blog', name: 'blog_list')]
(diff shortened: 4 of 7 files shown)


mcsky:
(i'm going to add a test for this)

config/sets/symfony/symfony64.php
```diff
@@ -12,6 +12,7 @@
         [
             'Symfony\Component\HttpKernel\UriSigner' => 'Symfony\Component\HttpFoundation\UriSigner',
             'Symfony\Component\HttpKernel\Debug\FileLinkFormatter' => 'Symfony\Component\ErrorHandler\ErrorRenderer\FileLinkFormatter',
+            'Symfony\Component\Routing\Annotation\Route' => 'Symfony\Component\Routing\Attribute\Route',
```
samsonasik, line 15: if `AnnotationToAttributeRector` is not called early, it may cause unexpected result, since it will still annotation, while renamed.

I think `RenameClassRector` need to be placed right **after** `AnnotationToAttributeRector` in the config annotation to attribute there

https://github.com/rectorphp/rector-symfony/blob/c9d78b386a7658e3f9ad71ad055a38eca436ee69/config/sets/symfony/annotations-to-attributes.php#L14

composer.json
```diff
@@ -13,6 +13,7 @@
         "phpstan/phpstan-webmozart-assert": "^1.2",
         "phpunit/phpunit": "^10.3",
         "rector/phpstan-rules": "^0.6",
+        "rector/rector": "*",
```
samsonasik, line 16: This is not needed

config/sets/symfony/symfony64.php
samsonasik: On AnnotationToAtrributeRector, the

`new AnnotationToAttribute('Symfony\Component\Routing\Annotation\Route' , 'Symfony\Component\Routing\Attribute\Route')`

should auto rename

mcsky: I'm in the train myself, not the best place to work. ^^ Thanks to point me to the right way. Do you have recommandation to learn rector ecosystem ?

config/sets/symfony/annotations-to-attributes.php
```diff
@@ -18,4 +19,10 @@
         // @see Symfony 5.2+ https://github.com/symfony/doctrine-bridge/commit/02d2cf4743331e6b69ffd1d68e09b7e2dc417201#diff-1a16e2739e51eab000116d0542bd0226cea59a6d64711740ed7ce14769f95d1b
         new AnnotationToAttribute('Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity'),
     ]);
+    $rectorConfig->ruleWithConfiguration(
+        RenameClassRector::class,
+        [
```
samsonasik, line 24: It seems not needed, just update the 2nd arg of `AnnotationToAttribute` arg to `Symfony\Component\Routing\Attribute\Route` above on line 18 should auto change to target attribute with new name

```php
new AnnotationToAttribute('Symfony\Component\Routing\Annotation\Route', 'Symfony\Component\Routing\Attribute\Route'
```

test is needed tho

config/sets/symfony/symfony64.php
samsonasik: @TomasVotruba write a book about rector

https://tomasvotruba.com/book-detail/rector-the-power-of-automated-refactoring?slug=rector-the-power-of-automated-refactoring

config/sets/symfony/annotations-to-attributes.php
samsonasik: /cc  @scuben @JohJohan could you verify that "Symfony\\Component\\Routing\\**Attribute**\\Route" class exists in specific symfony version when upgrading from "Symfony\\Component\\Routing\\**Annotation**\\Route" thank you

mcsky: `Symfony\Component\Routing\Attribute\Route` is available since symfony 6.4 AFAIK

samsonasik: That's what I mean, if that target class introduced in later version, eg transforming annotation is start from symfony 5.2, while the new target attribute class only exists since 6.4, then that RenameClassRector need to be placed in specific setlists, the issue is, that need to be checked that rename attribute, not rename annotation

samsonasik:
Ok, I am checking, the PHP 8 route attribute support exists since symfony 5.2, but the new class: "Symfony\\Component\\Routing\\**Attribute**\\Route" not yet exists, ref:

- https://symfony.com/blog/new-in-symfony-5-2-php-8-attributes
- https://github.com/symfony/symfony/tree/5.2/src/Symfony/Component/Routing
- https://github.com/symfony/symfony/blob/7fa617d4175e08a3d8e1cb5d67aa4523ad034dd8/src/Symfony/Component/Routing/Annotation/Route.php#L23-L24

only the class: "Symfony\\Component\\Routing\\**Annotation**\\Route".


Then, start from symfony 6.4, the "Symfony\\Component\\Routing\\**Attribute**\\Route" exists,

https://github.com/symfony/symfony/blob/914c2880393fe31335e716809edfdf66beedd246/src/Symfony/Component/Routing/Attribute/Route.php#L26-L27

but the `Annotation` sub namespace is not deprecated, see 

https://github.com/symfony/symfony/blob/914c2880393fe31335e716809edfdf66beedd246/src/Symfony/Component/Routing/Annotation/Route.php#L12-L16

so the original "Symfony\Component\Routing\**Annotation**\Route" is still valid, which means the mapping is not necessarily needed.

If this is needed, to avoid invalid rename, as it only should be used in the attribute part, I think custom rule is needed which the step is:

- create custom rule
- verify if it is attribute in method, if match, rename
- register the rule to symfony 6.4 setlist

https://github.com/rectorphp/rector-symfony/blob/main/config/sets/symfony/symfony64.php 

which the "Symfony\\Component\\Routing\\**Attribute**\\Route" exists.

mcsky:
I give it a try very soon, thanks for all the details !

config/sets/symfony/symfony64.php
mcsky: I'll ask my company to buy it :)

rules/Symfony64/Rector/Class_/FixImportWhileUsingRouteAttributeButAnnotationImportRector.php
```diff
@@ -0,0 +1,74 @@
...
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if ($node instanceof Attribute && $node->name !== null && (string) $node->name === self::ANNOTATION_ROUTE) {
+            // there I'd like to modify the UseUse node representing my current node->name
+            // AFAIK the node I'm returning from this method is only the Attribute node
+            // maybe should I retrieve the global AST from somewhere and manually get the UseUse node to edit the value
+            // Or maybe there is a more elegant way to do that ?
+
+            return null;
```
mcsky, line 69: I would like some help on this part. I think this if statement is enough to identify that the #[Route] attribute is used in the file.

But I'm stuck to modify the related UseUse node of this attribute. Have you any recommendations on this ?

mcsky:
@samsonasik I started an implementation of the behaviour you described but I'm stuck. I can't figure out how to modify the use statement related to the #[Route] attribute node

rules/Symfony64/Rector/Class_/FixImportWhileUsingRouteAttributeButAnnotationImportRector.php
samsonasik: No need to import to use statement, that will be taken care by : `$rectorConfig->importNames()` when enabled.

rules/Symfony64/Rector/Class_/FixImportWhileUsingRouteAttributeButAnnotationImportRector.php
```diff
@@ -0,0 +1,74 @@
...
+use PhpParser\Node;
+use PhpParser\Node\Attribute;
+use Rector\Rector\AbstractRector;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+/**
+ * @see \Rector\Symfony\Tests\Symfony64\Rector\Class_\FixImportWhileUsingRouteAttributeButAnnotationImportRector
+ */
+final class FixImportWhileUsingRouteAttributeButAnnotationImportRector extends AbstractRector
```
samsonasik, line 16:
```suggestion
final class ChangeRouteAttributeFromAnnotationSubnamespaceRector extends AbstractRector
```

rules/Symfony64/Rector/Class_/FixImportWhileUsingRouteAttributeButAnnotationImportRector.php
samsonasik: Oh I see what you mean, in that case, `$rectorConfig->importNames()` is working on `RenameClassRector`, but not this custom rector, you may need to lookup of what AnnotationToAttributeRector implementation, set Name node, and add Name node to import nodes collector to be imported later

samsonasik: You can use `Rector\PostRector\Collector\UseNodesToAddCollector` service for that, eg:

```php
$this->useNodesToAddCollector->addUseImport(new FullyQualifiedObjectType($namespacedAttrName));
```

see https://github.com/rectorphp/rector-src/blob/663f526966313e0f178c088ea667119490404da9/rules/Php80/Rector/Property/NestedAnnotationToAttributeRector.php#L219

rules/Symfony64/Rector/Class_/FixImportWhileUsingRouteAttributeButAnnotationImportRector.php
```diff
@@ -0,0 +1,74 @@
...
+CODE_SAMPLE
+                ),
+
+            ]
+        );
+    }
+
+    public function getNodeTypes(): array
+    {
+        return [Attribute::class];
```
samsonasik, line 55: use `ClassMethod` instead, and rename its `attrGroups` so it only change under `ClassMethod`

mcsky: The #Route[] attribute can be specified on a method but also on a class in Symfony

samsonasik: then, you need to define in `Class_` and `ClassMethod`

rules/Symfony64/Rector/Class_/FixImportWhileUsingRouteAttributeButAnnotationImportRector.php
samsonasik: After some thinking, replace use statement directly is unnecessary on this case, if that needed, you can do:

```php
$rectorConfig->importNames();
$rectorConfig->removeUnusedImports();
```

That will be safer imo, and that will be replaced.

The improvement for this case need to be verified carefully, and that need separate PR.

rules-tests/Symfony64/Rector/Class_/ChangeRouteAttributeFromAnnotationSubnamespaceRector/Fixture/use-annotation-should-skip.php.inc
```diff
@@ -0,0 +1,41 @@
...
+     * @Route("/")
+     */
+    public function __invoke(Request $request): JsonResponse
+    {
+        return new JsonResponse();
+    }
+}
+
+?>
+-----
```
samsonasik, line 21: The second part start from `-----` can be removed if equal content

rules-tests/Symfony64/Rector/Class_/ChangeRouteAttributeFromAnnotationSubnamespaceRector/Fixture/use-annotation-should-skip.php.inc
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * @Route("/")
+     */
+    public function __invoke(Request $request): JsonResponse
+    {
+        return new JsonResponse();
+    }
+}
+
+?>
```
samsonasik, line 41:
```suggestion
```

samsonasik (approved):
Looks good, thank you

samsonasik:
Thank you @mcsky

TomasVotruba:
Thanks @mcsky 🙏

PR #572, opened 2024-01-17
Deprecate Symfony level sets

TomasVotruba:
It's not best practise to use `UP_TO_*` sets in `rector.php` longer than is their single-run. They run dozens of rules with very complex logic, as it take to get from Symfony 2 to 7. It's like running a Tesla control and checking all elements available since the first horse ride :) 

See section 3:
https://getrector.com/blog/5-common-mistakes-in-rector-config-and-how-to-avoid-them

It's best practice - and we use it this way since the first Symfony rule set - to run those just once and remove .

<br>

Instead use only the latest major Symfony sets, to keep up to date and get fast and reliable Rector run :+1: 

Thank you

---

* Resolves https://github.com/rectorphp/rector/issues/8403
* Resolves https://github.com/rectorphp/rector-symfony/pull/547


TomasVotruba:
See https://github.com/rectorphp/rector-src/pull/5477

PR #584, opened 2024-02-29
Fixes the MessageHandlerInterfaceToAttributeRector rule reported as always applied

carlos-granados:
The `checkForServices()` function should return `null` if it has not actually changed the class

This is part of the SYMFONY_62 rule set

### DIFF
--- a/rules/Symfony62/Rector/Class_/MessageHandlerInterfaceToAttributeRector.php
+++ b/rules/Symfony62/Rector/Class_/MessageHandlerInterfaceToAttributeRector.php
@@ -107,16 +107,19 @@ CODE_SAMPLE
      */
     private function checkForServices(Class_ $class, array $handlers): ?Class_
     {
+        $hasChanged = false;
         foreach ($handlers as $handler) {
             if ($this->isName($class, $handler->getClass() ?? $handler->getId())) {
                 $options = $this->messengerHelper->extractOptionsFromServiceDefinition($handler);
                 if (! isset($options['method']) || $options['method'] === '__invoke') {
                     $this->messengerHelper->addAttribute($class, $options);
-                    return $class;
+                    $hasChanged = true;
                 }
             }
         }
-
+        if ($hasChanged) {
+            return $class;
+        }
         return null;
     }
 }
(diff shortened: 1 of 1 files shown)


rules/Symfony62/Rector/Class_/MessageHandlerInterfaceToAttributeRector.php
```diff
@@ -105,17 +105,18 @@ public function refactor(Node $node): ?Node
...
      */
-    private function checkForServices(Class_ $class, array $handlers): Class_
+    private function checkForServices(Class_ $class, array $handlers): ?Class_
     {
         foreach ($handlers as $handler) {
             if ($this->isName($class, $handler->getClass() ?? $handler->getId())) {
                 $options = $this->messengerHelper->extractOptionsFromServiceDefinition($handler);
                 if (! isset($options['method']) || $options['method'] === '__invoke') {
                     $this->messengerHelper->addAttribute($class, $options);
+                    return $class;
```
samsonasik, line 115: is multiple `addAttribute()` can happen? if yes, the return can in after foreach, eg, via set `$hasChanged = false` before loop, and set `$hasChanged = true` on this line, then after foreach, verify `if ($hasChanged) { return $class; }`

carlos-granados: I don't think that you can have multiple `AsMessageHandler` attributes but not 100% sure so I implemented this change just in case

samsonasik:
Thank you @carlos-granados

rules/Symfony62/Rector/Class_/MessageHandlerInterfaceToAttributeRector.php
stefantalen: > A single handler class can handle multiple messages. For that add the `#AsMessageHandler` attribute to all the handling methods

https://symfony.com/doc/current/messenger.html#handling-multiple-messages

samsonasik: @stefantalen the usage of `hasChanged` flag is the safest way for that

PR #585, opened 2024-03-01
Update README to use the new config syntax

carlos-granados:
Use `RectorConfig::configure()` instead of `static function (RectorConfig $rectorConfig)`


samsonasik (approved):

samsonasik:
Thank you @carlos-granados

PR #586, opened 2024-03-04
Fixes namespace of `ChainRequestMatcher`

mvhirsch:
That file does not exist :sweat_smile: 

https://github.com/symfony/symfony/tree/6.2/src/Symfony/Component/HttpFoundation/RequestMatcher

It was introduced in https://github.com/symfony/symfony/pull/47595


samsonasik (approved):

samsonasik:
Thank you @mvhirsch

PR #589, opened 2024-03-11
EventListenerToEventSubscriberRector runs also if AsEventSubscriber-Attribute is present

CMH-Benny:
Add 2 fixtures to cover Listener-Classes using AsEventListener-Attributes that are going to be changed to EventSubscriber without any need

### DIFF
--- a/rules-tests/CodeQuality/Rector/Class_/EventListenerToEventSubscriberRector/Fixture/skip_when_as_event_listener_attribute_exists_on_class.php.inc
+++ b/rules-tests/CodeQuality/Rector/Class_/EventListenerToEventSubscriberRector/Fixture/skip_when_as_event_listener_attribute_exists_on_class.php.inc
@@ -9,19 +9,3 @@ class SingleMethodListener
     {
     }
 }
-
-?>
------
-<?php
-
-namespace Rector\Symfony\Tests\CodeQuality\Rector\Class_\EventListenerToEventSubscriberRector\Fixture;
-
-#[AsEventListener]
-class SingleMethodListener
-{
-    public function __invoke()
-    {
-    }
-}
-
-?>
--- a/rules-tests/CodeQuality/Rector/Class_/EventListenerToEventSubscriberRector/Fixture/skip_when_as_event_listener_attribute_exists_on_class.php.inc
+++ b/rules-tests/CodeQuality/Rector/Class_/EventListenerToEventSubscriberRector/Fixture/skip_when_as_event_listener_attribute_exists_on_class.php.inc
@@ -2,6 +2,8 @@
 
 namespace Rector\Symfony\Tests\CodeQuality\Rector\Class_\EventListenerToEventSubscriberRector\Fixture;
 
+use Symfony\Component\EventDispatcher\Attribute\AsEventListener;
+
 #[AsEventListener]
 class SingleMethodListener
 {
--- a/rules-tests/CodeQuality/Rector/Class_/EventListenerToEventSubscriberRector/Fixture/skip_when_as_event_listener_attribute_exists_on_class.php.inc
+++ b/rules-tests/CodeQuality/Rector/Class_/EventListenerToEventSubscriberRector/Fixture/skip_when_as_event_listener_attribute_exists_on_class.php.inc
@@ -1,0 +1,27 @@
+<?php
+
+namespace Rector\Symfony\Tests\CodeQuality\Rector\Class_\EventListenerToEventSubscriberRector\Fixture;
+
+#[AsEventListener]
+class SingleMethodListener
+{
+    public function __invoke()
+    {
+    }
+}
+
+?>
+-----
+<?php
+
+namespace Rector\Symfony\Tests\CodeQuality\Rector\Class_\EventListenerToEventSubscriberRector\Fixture;
+
+#[AsEventListener]
+class SingleMethodListener
+{
+    public function __invoke()
+    {
+    }
+}
+
+?>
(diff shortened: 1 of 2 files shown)


CMH-Benny:
Yeah, it's a little confusing for me, I was cleaning a project that was using Listeners and Rector wanted to convert to EventSubscriber, I wondered why I did some digging and eded up creating this issue on the main repo: https://github.com/rectorphp/rector/issues/8532#issue-2175215591

If I follow the reasoning of the conversion to EventSubscriber it was because it doesn't need any additional config, but can be done in the EventSubscriber class - However with the Attribute this also applies to Listeners now and I would assume the Attribute Way even more modern, than the method returning an array within an array to register your event/method/priority - So for me the AsListener seems to be superior now. Not sure if there is a case where you still want/have to use the EventSubscriber, tho

rules-tests/CodeQuality/Rector/Class_/EventListenerToEventSubscriberRector/Fixture/skip_when_as_event_listener_attribute_exists_on_class.php.inc
```diff
@@ -0,0 +1,27 @@
...
+
+#[AsEventListener]
+class SingleMethodListener
+{
+    public function __invoke()
+    {
+    }
+}
+
+?>
```
samsonasik, line 27: second part is not needed for skip fixture

```suggestion
```

rules-tests/CodeQuality/Rector/Class_/EventListenerToEventSubscriberRector/Fixture/skip_when_as_event_listener_attribute_exists_on_methods.php.inc
```diff
@@ -0,0 +1,37 @@
...
+    {
+    }
+
+    #[AsEventListener]
+    public function onMethodB()
+    {
+    }
+}
+
+?>
```
samsonasik, line 37: second part is not needed for skip fixture

```suggestion
```

CMH-Benny: I pushed the changes

CMH-Benny:
Any update on this?

stefantalen:
Converting an EventListener to an EventSubscriber and replacing EventListeners with `#[AsEventListener]` are two seperate things (with their own use cases).

I'll see if I can wrap up my PR within a week to include:
- Remove `EventListenerToEventSubscriberRector` from `SYMFONY_CODE_QUALITY` set
- Extend `EventListenerToEventSubscriberRector` to also convert to EventSubscriber, even if `AsEventListener` attribute is present for anyone who want to use it

It'll be up to the developer to include the rule or not 🙂

Maybe you can work on a `EventListenerToAsEventListenerAttributeRector`?

I don't think it should be part of a Symfony set, since using the attribute is still optional

CMH-Benny:
Thanks for your answer, that sounds okay to me. However I am not sure if it should really be dropped from the CODE_QUALITY, the idea of it was to ensure people don't use the old way with service tags somewhere configured in a config file, but to use the EventSubscriberInterface to have the config basically in the same class as the Subscriber. However, for the new AsEventListener Attribute this is no longer special to EventSubscriber, since the Attribute basically also puts the config inthe the Listener class. So there are 2 solutions that would fit into that CODE_QUALITY Set, AsEventListener-Attribute approach and the EventSubscriberInterface, so not sure which one is the better pick for that Set.

I lately migrated a symfony 5.4 to a symfony 6.4 migration and Presets for SYMFONY_6{0-4} started to convert MessageHandlers, Commands etc. to use Attributes, so such rools are part of the symfony rules sets already, for some reason it doesn't apply to EventListeners (yet) so on one side I think it's not a good idea, to drop that EventSubscriberRector from the CodeQuality, however I also highly doubt that it should revert back AsEventListener Attributes into EventSubscriberInterface, but I am also not sure, what is considered best practice now, as stated we have two conflicting soutions that fulfil both the reasoning why they should be in the CODE_QUALITY set. 

I for myself disabled/skipped the EventSubscriberRector so it doesn't destroy my Attributes, I don't see a reason why I would want to use the Interface over the Attribute, especially since Rector actively changed towards Attribute in other cases. I just wanted to bring that topic to attention, sadly I won't have time to dig into the creation of rector rules for a while, so I am fine with every solution. If you are going to convert to EventSubscriber even if people use attributes, people can at any time decide to disable that rector as I did and maybe at some point the AsEventListener-Attribute becomes the best practice and Rector will adjust towards it

Thanks for your efforts, rector is really an amazing tool and if I ever get some time to dig deeper into this whole thing, I'd be happy to contribute at some point =)

TomasVotruba:
Hi, thanks for patience with this. I was focusing on other issues, but I'm giving this priority now.

Goal of this rule is to move configuration logic from some config out there, directly to PHP code.
Both `EventSubscriberInterface` and `#[AsEventListener]` make it happen.

https://symfony.com/doc/current/event_dispatcher.html#event-dispatcher_event-listener-attributes

```php
#[AsEventListener(event: CustomEvent::class, method: 'onCustomEvent')]
#[AsEventListener(event: 'foo', priority: 42)]
#[AsEventListener(event: 'bar', method: 'onBarEvent')]
final class MyMultiListener
```

---

So we should go with:

> Ignore when attribute is present

I'll process this today. Thanks again for your patience :pray:

TomasVotruba:
I've cherry picked your fixtures in https://github.com/rectorphp/rector-symfony/pull/616

Thanks :+1:

CMH-Benny:
Thank you so much! 
I will delete my branch then, happy to hear that my fixtures helped you =)

PR #591, opened 2024-04-04
Add missing 'Serializer/Context' class to the list of "annotation to attribute" rule

julienfastre:
This  MR add a missing class to the list of configured "annotation to attributes" set list: the class `Symfony\Component\Serializer\Annotation\Context`, which was not yet configured.


TomasVotruba:
Thank you :+1:

PR #593, opened 2024-04-05
Allow null values for domain and locale in Translator

### LINKED ISSUE #592: The rector rule for TranslatorInterface does not take into account possible null parameter
The signature of [`Symfony\Contracts\Translation\TranslatorInterface::trans`](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Contracts/Translation/TranslatorInterface.php) is 

```php
interface TranslatorInterface
{
    // ...

    public function trans(string $id, array $parameters = [], ?string $domain = null, ?string $locale = null): string;

   // ...
}

```

But the configured rules does not add the `?` before the `?string $domain = null`:

```
    ---------- begin diff ----------
@@ @@
     public function getFilter()
     {
         $translator = new class () implements TranslatorInterface {
-            public function trans(string $id, array $parameters = [], ?string $domain = null, string $locale = null)
+            public function trans(string $id, array $parameters = [], string $domain = null, string $locale = null)
             {
                 return $id;
             }
    ----------- end diff -----------

Applied rules:
 * AddParamTypeDeclarationRector
```

As it is syntactically correct to remove this `?`, it leads to conflicts between rector rules and other tools, like php-cs-fixer.

Can I suggest a PR to add this question mark ?

julienfastre:
The 'trans' method in the Translator class has been updated to allow null values for 'domain' and 'locale' parameters.

Closes #592

### DIFF
--- a/config/sets/symfony/symfony50-types.php
+++ b/config/sets/symfony/symfony50-types.php
@@ -267,13 +267,13 @@ return static function (RectorConfig $rectorConfig): void {
             'Symfony\Contracts\Translation\TranslatorInterface',
             'trans',
             2,
-            new StringType()
+            \PHPStan\Type\TypeCombinator::addNull(new StringType())
         ),
         new AddParamTypeDeclaration(
             'Symfony\Contracts\Translation\TranslatorInterface',
             'trans',
             3,
-            new StringType()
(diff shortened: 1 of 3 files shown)


TomasVotruba:
Thank you :+1:

PR #595, opened 2024-04-09
Fix TemplateGuesser to snake case controller classes and actions

MikeAmelung:
Currently, TemplateAnnotationToThisRenderRector is leaving the action method names camel cased. It was also leaving the controller capitalized with underscores. I believe this fix makes it work according to the docs at: https://symfony.com/bundles/SensioFrameworkExtraBundle/current/annotations/view.html

### DIFF
--- a/src/Helper/TemplateGuesser.php
+++ b/src/Helper/TemplateGuesser.php
@@ -89,16 +89,22 @@ final readonly class TemplateGuesser
 
         $action = Strings::replace($method, self::ACTION_MATCH_REGEX, '');
 
+        $action = Strings::replace(
+            $action,
+            self::SMALL_LETTER_BIG_LETTER_REGEX,
+            '\\1_\\2'
+        );
+
         $fullPath = '';
         if ($bundle !== '') {
             $fullPath .= $bundle . '/';
         }
 
         if ($controller !== '') {
-            $fullPath .= $controller . '/';
+            $fullPath .= strtolower($controller) . '/';
         }
 
-        return $fullPath . $action . '.html.twig';
+        return $fullPath . strtolower($action) . '.html.twig';
     }
 
     private function resolveBundle(string $class, string $namespace): string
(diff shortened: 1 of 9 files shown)


TomasVotruba:
Looks good, thank you :+1:

src/Helper/TemplateGuesser.php
```diff
@@ -89,16 +89,22 @@ private function resolve(string $namespace, string $class, string $method): stri
 
         $action = Strings::replace($method, self::ACTION_MATCH_REGEX, '');
 
+        $action = Strings::replace(
+            $action,
+            self::SMALL_LETTER_BIG_LETTER_REGEX,
+            '\\1_\\2'
```
samsonasik, line 95: this somehow changed to `1_\\2` in scoped rector/rector https://github.com/rectorphp/rector/commit/67791a95820367d299e8e891b18d91afc9fb08d5#diff-02557c06605d556a23eae70c6c7b37cf26cc8f54ca434718889e7a97d8d8a2e3

@MikeAmelung probably changed by php-scoper, could you verify latest rector/rector?

MikeAmelung: I can confirm that it is broken in rectorphp/rector, as that is what initially caught my attention as my controllers were coming out as "Som1_Controller". I thought to mention it in my original pull request comment, but I didn't realize it was a build process issue causing it.

samsonasik: any alternative code for it? If no, the scoper.php may need str_replace for this file, eg:

https://github.com/rectorphp/rector-src/blob/1d2d51e20ce11d0f603e42bd8869c1ab6d2a52cc/scoper.php#L82-L89

MikeAmelung: Added a new commit, hopefully this doesn't cause any issues with the scoper? I had copied the '\\1_\\2' from code that was already there, but I feel like '$1_$2' is more common usage with regexes anyway.

https://github.com/MikeAmelung/rector-symfony/commit/1f61d849decfaa1a624482aaf0c16fe065a79e46

Should I create another pull request?

samsonasik: Yes please :+1:

PR #596, opened 2024-04-09
Replace \ with $ for capturing group reference in TemplateGuesser

MikeAmelung:
Fix to get past the scoper


samsonasik:
Thank you @MikeAmelung

MikeAmelung:
Confirmed to be working on rectorphp/rector@dev-main now. Thank you and @TomasVotruba and everyone, glad I could contribute.

PR #603, opened 2024-05-27
Convert \Swift_Message to an Symfony\Component\Mime\Email

stefantalen:


TomasVotruba:
Thank you :+1:

rules/SwiftMailer/Rector/ClassMethod/Swift_MessageToEmailRector.php
```diff
@@ -0,0 +1,232 @@
...
+                }
+                $args = $node->getArgs();
+                if (count($args) > 0) {
+                    $node = new MethodCall(new New_(new FullyQualified(self::EMAIL_FQN)), 'subject', [$args[0]]);
+                } else {
+                    $node->class = new FullyQualified(self::EMAIL_FQN);
+                }
+            }
+
+            if ($node instanceof MethodCall) {
```
stefantalen, line 124: @TomasVotruba Just noticed a bug, unrelated method calls are being updated as well.

I tried adding
```php
if (!$this->isObjectType($node->var, new ObjectType(self::SWIFT_MESSAGE_FQN))) {
    return $node;
}
```
here, but that triggers a failure on daisy chained methods because `$node->var` is a MethodCall. Any suggestions?

samsonasik: @stefantalen please create failing test case PR for skip fixture so we can take a look

stefantalen: #604

PR #607, opened 2024-05-30
Enable rectorPreset set

samsonasik:
see https://github.com/rectorphp/rector-src/pull/5924

### DIFF
--- a/rules/Configs/Rector/Closure/ServiceSettersToSettersAutodiscoveryRector.php
+++ b/rules/Configs/Rector/Closure/ServiceSettersToSettersAutodiscoveryRector.php
@@ -105,7 +105,7 @@ CODE_SAMPLE
         $classNamesAndFilesPaths = $this->createClassNamesAndFilePaths($bareServicesSetMethodCallExpressions);
 
         $classNames = array_map(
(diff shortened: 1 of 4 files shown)


rules/Configs/Rector/Closure/ServiceSettersToSettersAutodiscoveryRector.php
```diff
@@ -105,7 +105,7 @@ public function refactor(Node $node): ?Node
         $classNamesAndFilesPaths = $this->createClassNamesAndFilePaths($bareServicesSetMethodCallExpressions);
 
         $classNames = array_map(
-            fn (ClassNameAndFilePath $classNameAndFilePath): string => $classNameAndFilePath->getClassName(),
+            static fn(ClassNameAndFilePath $classNameAndFilePath): string => $classNameAndFilePath->getClassName(),
```
TomasVotruba, line 108: Oh, I've missed this in Rector pre-set. I want to have it removed as well, same for closures. It makes code much less readable for me. Thanks :+1:

samsonasik: I personally prefer use static, as static closure/arrowfunction is faster, but sure, here it is PR to remove that:

- https://github.com/rectorphp/rector-src/pull/5927

Issue #598, opened 2024-04-18
Bug: ContainerGetToConstructorInjectionRector replaces constant with object


bbrala:
I've seen this issue also while running against the Drupal codebase. I've got a reproducable test :) 

https://getrector.com/demo/0c616439-8d54-46f3-925c-73aefd8ce429

Reopens #439

bbrala:
I've added a PR that fixes this issue @TomasVotruba  :)

PR #610, opened 2024-06-18
Fix constr fetch resulting in invalid injection

bbrala:
Fixes: https://github.com/rectorphp/rector-symfony/issues/598

Proof of error: https://getrector.com/demo/b7387a86-a969-49c2-9ef5-3bf163f6e036

### DIFF
--- a/rules-tests/Symfony42/Rector/MethodCall/ContainerGetToConstructorInjectionRector/Fixture/constant_as_service_regression.php.inc
+++ b/rules-tests/Symfony42/Rector/MethodCall/ContainerGetToConstructorInjectionRector/Fixture/constant_as_service_regression.php.inc
@@ -1,0 +1,213 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\g2\Controller;
+
+use Drupal\Core\Cache\CacheableJsonResponse;
+use Drupal\Core\Cache\CacheableMetadata;
+use Drupal\Core\Config\ConfigFactoryInterface;
+use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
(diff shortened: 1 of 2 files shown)


rules-tests/Symfony42/Rector/MethodCall/ContainerGetToConstructorInjectionRector/Fixture/constant_as_service_regression.php.inc
```diff
@@ -0,0 +1,213 @@
...
+        /** @var \Drupal\g2\RouteFilter $rf */
+        $rf = $container->get(G2::SVC_ROUTE_FILTER);
+
+        /** @var \Drupal\g2\WOTD $wotd */
+        $wotd = $container->get(G2::SVC_WOTD);
+
+        return new static($entity_manager, $rf, $wotd, $configFactory);
+    }
+}
+?>
```
TomasVotruba, line 213: Thanks for the fix :+1: 

Could you make this test fixture as small as possible? So we can read it easily and see the buggy part :)

bbrala: Yeah can do later today. Should be easy enough

bbrala:
I've updated the test to a minimal test that prooves the issue.

rules-tests/Symfony42/Rector/MethodCall/ContainerGetToConstructorInjectionRector/Fixture/constant_as_service_regression.php.inc
bbrala: Minized the test

samsonasik (approved):

Issue #537, opened 2023-10-23
Symfony 7 rector rule collection


alexander-schranz:
I just want to create this issue to collect upgrade rule ideas I stumble over when working on supporting Symfony 7 for some packages.

- [x] classes implementing `Symfony\Component\Form\DataTransformerInterface` requires return types

```diff
-    public function transform($document)
+    public function transform($document): ?string

-    public function reverseTransform($uuid)
+    public function reverseTransform($uuid): ?object
```

----

to be continued ...

----

The question mostly will be should they already be part of the 6.4 Symfony ruleset. When used by a library it could something like adding a return type be a backwards compatibility break. But when used in a project it would not hurt to have the required return types already there, so not sure how this is currently handled by rector rulesets.

stefantalen:
Looking at the current sets (5.0 and 6.0), there a separate sets (`config/sets/symfony/symfony50-types.php` and `config/sets/symfony/symfony6/symfony-return-types.php`) with the type changes which is imported in the actual version set.

Not sure with the reasoning behind it, but probably for cleaner sets. 🙂

aragon999:
Something I just noticed when upgrading a project to Symfony 7, was that `Symfony\Component\Routing\Annotation\Route` -> `Symfony\Component\Routing\Attribute\Route`.

Not sure if this should be contained in the Symfony 6.4 set or Symfony 7 set.

samsonasik:
Its already in symfony 6.4 set, see PR:

- https://github.com/rectorphp/rector-symfony/pull/557

Try latest rector/rector:dev-main

aragon999:
Thank you :-)

TomasVotruba:
I'm looking into this...

TomasVotruba:
@alexander-schranz Thanks for kick of :)

The data transformer return types added in https://github.com/rectorphp/rector-symfony/pull/624

> The question mostly will be should they already be part of the 6.4 Symfony ruleset. When used by a library it could something like adding a return type be a backwards compatibility break.

This is the best choice, as Symfony would add types already in latest major version, 6.4, so extending would comply already with Symfony 6.4 :+1:

TomasVotruba:
> Not sure with the reasoning behind it, but probably for cleaner sets. 🙂

Indeed, there were so many changes in Symfony 5.0 we decided to separate a new set. Just to have better control over changed during an upgrade :)

Return type changes in 6.4 are actually quite few, so we can have it in one set.

[...]

PR #624, opened 2024-06-22
[Symfony 6.4] Add DataTransformerInterface return types

TomasVotruba:
Ref https://github.com/rectorphp/rector-symfony/issues/537

### DIFF
--- a/config/sets/symfony/symfony64.php
+++ b/config/sets/symfony/symfony64.php
@@ -2,10 +2,15 @@
 
 declare(strict_types=1);
 
+use PHPStan\Type\ObjectWithoutClassType;
+use PHPStan\Type\StringType;
+use PHPStan\Type\UnionType;
 use Rector\Config\RectorConfig;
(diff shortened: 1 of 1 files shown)


trsteel88:
@TomasVotruba is this forcing a return type on DataTransformers in symfony? 

I have the following

```php
public function reverseTransform(mixed $value): ?string
    {
        if (null === $value || '' === (string) $value) {
            return null;
        }

        return 'foo';
    }
```

But rector is changing this to:

```php
public function reverseTransform(mixed $value): object|string
    {
        if (null === $value || '' === (string) $value) {
            return null;
        }

        return 'foo';
    }
```

But this isn't correct. It can only be `?string` so now I have PHPStan failing on me:

```
::reverseTransform() should return object|string but returns null.  
```

trsteel88:
Ah using the 71 set instead of 64 fixed it.

`$rectorConfig->import(Rector\Symfony\Set\SymfonySetList::SYMFONY_71);`

PR #632, opened 2024-07-01
Hybrid CommandConfigureToAttributeRector

ruudk:
When you already have a AsCommand attribute, but still have things that are done in `configure`, move them over.


TomasVotruba:
Looks good, thank you 👍

Issue #393, opened 2023-04-18
[Symfony62] Replace Sensio Security


wkania:
[sensio-framework-extra-bundle](https://packagist.org/packages/sensio/framework-extra-bundle) is abandoned.
Some attributes have been implemented in Symfony 6.2 and Rector can already refactor code for [them](https://github.com/rectorphp/rector-symfony/pull/265).
But, `Security` attribute will not be [implemented](https://github.com/symfony/symfony/pull/45415#issuecomment-1077625813) in Symfony.  
From what I know about Rector we can't currently make change [from](https://symfony.com/bundles/SensioFrameworkExtraBundle/current/annotations/security.html):
```
use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;

class PostController extends Controller
{
    #[Security("is_granted('ROLE_ADMIN') and is_granted('ROLE_FRIENDLY_USER')")]
    public function index()
    {
    }
}
```
[to](https://symfony.com/doc/current/security/expressions.html):
```
use Symfony\Component\ExpressionLanguage\Expression;
use Symfony\Component\Security\Http\Attribute\IsGranted;

class PostController extends Controller
{
    #[IsGranted(new Expression("is_granted('ROLE_ADMIN') and is_granted('ROLE_FRIENDLY_USER')"))]
    public function index()
    {
    }
}
```
This require a new rule, something custom like [this](https://github.com/rectorphp/rector-phpunit/pull/133)?

TomasVotruba:
Yes, this will need a custom rule, to handle the expressions.

johanadivare:
Well its good to have these 2 cases in testfixure

pepeh:
@wkania 
Why not replace
```
#[Security("is_granted('ROLE_FRIENDLY_USER', author'])")]
public function showAction(User $author): Response
```

with
```
#[IsGranted('ROLE_FRIENDLY_USER', 'author')]
public function showAction(User $author): Response
```

wkania:
@pepeh 
Yes, you are right. My example was to simple.
In my opinion, these are the valid cases and probably many [more](https://symfony.com/doc/current/security/expressions.html):
   
    -------------------
    #[Security("is_granted('ROLE_ADMIN')")]
    #[IsGranted('ROLE_ADMIN')]

    -------------------
    #[Security("is_granted('ASSET_EDIT', asset)")]
    #[IsGranted('ASSET_EDIT', 'asset')]

    -------------------
[...]

PR #583, opened 2024-02-14
[Symfony 4.4] Updating AuthorizationCheckerIsGrantedExtractorRector

stefantalen:
Felt like this rule was missing the AbstractController implementation.

This PR will add `AbstractController::isGranted()` ~and `AbstractController::denyAccessUnlessGranted()`~ support

PR #640, opened 2024-07-04
[Symfony 6.2] Improve SecurityAttributeToIsGrantedAttributeRector

TomasVotruba:
Follow up to https://github.com/rectorphp/rector-symfony/pull/622

Covered cases:

```diff
-#[Security("is_granted('ROLE_ADMIN')")]
+#[\Symfony\Component\Security\Http\Attribute\IsGranted(attribute: 'ROLE_ADMIN')]
```

```diff
-#[Security("is_granted('ROLE_ADMIN', someArea)")]
+#[\Symfony\Component\Security\Http\Attribute\IsGranted(attribute: 'ROLE_ADMIN', subject: 'someArea')]
```

I've looked at the other cases, but due to it's stringy nature, we can't parse it and covert it into different structures.
This is the best we can automate, the rest would have to be dealt with manually.

<br>

Closes https://github.com/rectorphp/rector-symfony/issues/393

### DIFF
--- a/rules-tests/Symfony62/Rector/Class_/SecurityAttributeToIsGrantedAttributeRector/Fixture/single_is_granted.php.inc
+++ b/rules-tests/Symfony62/Rector/Class_/SecurityAttributeToIsGrantedAttributeRector/Fixture/single_is_granted.php.inc
@@ -1,0 +1,31 @@
+<?php
+
+namespace Rector\Symfony\Tests\Symfony62\Rector\Class_\SecurityAttributeToIsGrantedAttributeRector\Fixture;
+
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
+
+final class SingleIsGranted
+{
+    #[Security("is_granted('ROLE_ADMIN')")]
+    public function index()
+    {
+    }
+}
+
+?>
+-----
+<?php
+
+namespace Rector\Symfony\Tests\Symfony62\Rector\Class_\SecurityAttributeToIsGrantedAttributeRector\Fixture;
+
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
+
+final class SingleIsGranted
+{
+    #[\Symfony\Component\Security\Http\Attribute\IsGranted('ROLE_ADMIN')]
+    public function index()
+    {
+    }
+}
+
+?>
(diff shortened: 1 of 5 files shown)


TomasVotruba:
cc @wkania I think this is the best we can get :)

TomasVotruba:
I've noticed array of roles in the Symofny PR: https://github.com/symfony/symfony/pull/46907

Would this help? Is this correct conversion?

```diff
-#[Security("is_granted('ROLE_ADMIN') or is_granted('ROLE_FRIENDLY_USER')")]
+#[IsGranted(attributes: ['ROLE_ADMIN', 'ROLE_FRIENDLY_USER'])]
```

wkania:
> I've looked at the other cases, but due to it's stringy nature, we can't parse it and covert it into different structures.

From the code I see that we do have access to this value so we can modify it? Is parsing complexity the problem?

> Would this help? Is this correct conversion?

There are a lot cases with or/and etc so I would be against such changes.

stefantalen:
> I've noticed array of roles in the Symofny PR: https://github.com/symfony/symfony/pull/46907
> 
> 
> 
> Would this help? Is this correct conversion?
> 
> 
> 
> ```diff
> 
> -#[Security("is_granted('ROLE_ADMIN') or is_granted('ROLE_FRIENDLY_USER')")]
> 
> +#[IsGranted(attributes: ['ROLE_ADMIN', 'ROLE_FRIENDLY_USER'])]
> 
> ```

Using arrays for roles has been deprecated, from what I know each role needs its own #[IsGranted]

I've recently made some changes in a rule for that

TomasVotruba:
> Using arrays for roles has been deprecated, from what I know each role needs its own #[IsGranted]
> I've recently made some changes in a rule for that

What kind of changes? Maybe we can includ them here.

TomasVotruba:
@wkania 

> From the code I see that we do have access to this value so we can modify it? Is parsing complexity the problem?

The parsing is not possible, because it's not PHP. It's only regex on string :) e.g: https://github.com/rectorphp/rector-symfony/pull/640/files#diff-56e5b788455db473092a5cc9b737f8301fe89a7c60863fd483d0fc71a84e8049R51

stefantalen:
https://github.com/rectorphp/rector-symfony/issues/583

TomasVotruba:
@stefantalen I see. That's probably out of scope of this attribute, right?

TomasVotruba:
The Attribute seems to support multiple roles, at least during it's first PR :)

https://github.com/symfony/symfony/pull/46907/files#diff-60f77e573fc91a3cfc2540e301a990850e2de17216050c55ce1d8b0aaeeccd58R16

TomasVotruba:
I think without before/after tested in the wild, there would be lot of guessing how these work.

Merging as this best we can get at the moment :)

PR #645, opened 2024-07-08
Correct return type definition for DataTransformerInterface

bobvandevijver:
Corrects the problem introduced #624 and fixes https://github.com/rectorphp/rector/issues/8730.

The correct return type is mixed (see https://github.com/symfony/symfony/blob/7.0/src/Symfony/Component/Form/DataTransformerInterface.php), the comment in https://github.com/rectorphp/rector-symfony/issues/537#issue-1957058803 where the change was based on was just an example for a particular scenario.

### DIFF
--- a/config/sets/symfony/symfony64.php
+++ b/config/sets/symfony/symfony64.php
@@ -2,10 +2,7 @@
 
 declare(strict_types=1);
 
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Thanks for prompt fix 🙏

Issue #646, opened 2024-07-12
StringToArrayArgumentProcessRector with $process->setInput()


refsz:
I have the following Code in my project:

` $process = new Process($command);
if (\in_array('pandoc', $command, true)) {
  $process->setInput('<p></p>');
}` 

The rule StringToArrayArgumentProcessRector changes the code for setInput to
`$process->setInput(['<p></p>']);`

Then PHPStan complains because setInput does not support arrays.

Should setInput be included in EXCLUDED_PROCESS_METHOD_CALLS?

samsonasik:
Could you provide a patch? the steps:

- add failing fixture into https://github.com/rectorphp/rector-symfony/tree/main/rules-tests/Symfony42/Rector/New_/StringToArrayArgumentProcessRector/Fixture , better add `@see` reference to symfony that pass array is invalid
- provide a patch

thank you.

PR #647, opened 2024-07-13
Do not convert to array Process->setInput()

refsz:
Corrects the problem introduced #646 

Closes https://github.com/rectorphp/rector-symfony/issues/646


samsonasik:
I think definition should be the other way around, instead of define the excluded method, define the will be processed method instead, so override new method will not automatically transformed

samsonasik:
I verified that pass string on `setInput()` will got error:

```
 "Symfony\Component\Process\Process::setInput" only accepts strings, Traversable objects or stream resources. 
```

samsonasik (approved):

samsonasik:
Thank you @refsz

PR #651, opened 2024-07-30
Add new Response() support to ResponseReturnTypeControllerActionRector

TomasVotruba:


rules/CodeQuality/Rector/ClassMethod/ResponseReturnTypeControllerActionRector.php
```diff
@@ -172,6 +186,48 @@ private function refactorResponse(ClassMethod $classMethod): Node
...
+        }
+
+        $returnedTypes = [];
+        foreach ($returns as $return) {
+            if (! $return->expr instanceof Node\Expr) {
+                // already validated above
+                throw new ShouldNotHappenException();
+            }
+
+            $returnedTypes[] = $this->getType($return->expr);
```
samsonasik, line 211: better to use native type on this case, ensuring no typo by docblock definition:

```suggestion
            $returnedTypes[] = $this->nodeTypeResolver->getNativeType($return->expr);
```

TomasVotruba: This is also one of cases, where we are more tolerant as return type comes from framework itself.

* Controller, with `#[Route()]` or `@Route()`, with response return type.

rules/CodeQuality/Rector/ClassMethod/ResponseReturnTypeControllerActionRector.php
```diff
@@ -144,34 +162,90 @@ private function hasReturn(ClassMethod $classMethod): bool
...
+
+        $returns = $this->betterNodeFinder->findReturnsScoped($classMethod);
+        if (! $this->returnAnalyzer->hasOnlyReturnWithExpr($classMethod, $returns)) {
+            return null;
+        }
+
+        // no return, no type
+        if ($returns === []) {
+            return null;
+        }
```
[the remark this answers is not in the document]
samsonasik: https://github.com/rectorphp/rector-symfony/pull/652

PR #659, opened 2024-08-08
Add the Symfony 7.0 set to SymfonySetList.php

pbowyer:
Every other Symfony release has its set listed, so this looks like an oversight?


samsonasik (approved):

samsonasik:
Thank you @pbowyer

PR #660, opened 2024-08-12
[CodeQuality] Skip posible return array + response on ResponseReturnTypeControllerActionRector

samsonasik:
ref https://github.com/rectorphp/rector-src/pull/6200#pullrequestreview-2207245293

it currently got incorrect return:

```diff
-    public function editAction(Request $request, $id = null)
+    public function editAction(Request $request, $id = null): ?array

Applied rules:
 * ResponseReturnTypeControllerActionRector
```

This PR fix it.


samsonasik:
Fixed 🎉

samsonasik:

samsonasik:
@TomasVotruba I am merging it to have faster feedback to test ;)

samsonasik:
I found another many use cases that it got invalid `?array` if the `Response` return is from custom method non-native return, I will add only return when natively

TomasVotruba:
Looks good :+1:

PR #681, opened 2024-11-19
Use rector-src#dev-upgrade-to-php-parser5-and-phpstan-2

samsonasik:
Ref:

- https://github.com/rectorphp/rector-src/pull/6431

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,11 +9,9 @@
     },
     "require-dev": {
         "phpstan/extension-installer": "^1.4",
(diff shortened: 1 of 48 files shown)


samsonasik:
@TomasVotruba finally, unit test green 🎉 🎉 🎉

now go to another `rector-*` package :)

TomasVotruba:
Looks good, let's ship this one as well :hugs:  Thank you!

PR #685, opened 2024-12-01
[DependencyInjection] Extract GetByTypeMethodCallToConstructorInjectionRector to make migration smoother

TomasVotruba:


rules/DependencyInjection/Rector/Class_/ControllerGetByTypeToConstructorInjectionRector.php
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+CODE_SAMPLE
+                    ,
+                    <<<'CODE_SAMPLE'
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+
+final class SomeCommand extends Controller
+{
+    public function __construct(private SomeType $someType)
```
samsonasik, line 56: min php version 8.0 is needed for promotion property

samsonasik: I see, it already handled at `$this->classDependencyManipulator->addConstructorDependency($node, $propertyMetadata)` that verify php version 👍

Issue #691, opened 2024-12-06
Could not process file, due to UrlGeneratorInterface not found.


biozshock:
Hello.

First of all thank you for the tool! Amazing.

Want to report a bug, where, rector can't seem to find it's bundled class `RectorPrefix202411\Symfony\Component\Routing\Generator\UrlGeneratorInterface`.

Stack trace:
```
[ERROR] Could not process "/var/www/code/rector-class-not-found/src/Check.php" file, due to:                           
         "System error: "Class "RectorPrefix202411\Symfony\Component\Routing\Generator\UrlGeneratorInterface" not found"
                                                                                                                        
         Stack trace:                                                                                                   
         #0                                                                                                             
         vendor/rector/rector/vendor/rector/rector-symfony/rules/Symfony26/Rector/MethodCall/RedirectToRouteRector.php(7
         4): Rector\Symfony\Symfony26\Rector\MethodCall\RedirectToRouteRector->isDefaultReferenceType()                 
         #1 vendor/rector/rector/src/Rector/AbstractRector.php(140):                                                    
         Rector\Symfony\Symfony26\Rector\MethodCall\RedirectToRouteRector->refactor()                                   
         #2 vendor/rector/rector/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(114):                          
         Rector\Rector\AbstractRector->enterNode()                                                                      
         #3 vendor/rector/rector/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(198):                          
         PhpParser\NodeTraverser->traverseNode()                                                                        
         #4 vendor/rector/rector/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(105):                          
         PhpParser\NodeTraverser->traverseArray()                                                                       
         #5 vendor/rector/rector/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(198):                          
         PhpParser\NodeTraverser->traverseNode()                                                                        
         #6 vendor/rector/rector/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(105):                          
         PhpParser\NodeTraverser->traverseArray()                                                                       
         #7 vendor/rector/rector/vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php(198):                          
         PhpParser\NodeTraverser->traverseNode()                                                                        
[...]

PR #692, opened 2024-12-06
[scoped] Fix scoped prefixed UrlGeneratorInterface on RedirectToRouteRector

samsonasik:
Fixes https://github.com/rectorphp/rector-symfony/issues/691


samsonasik:
Fixed 🎉 /cc @biozshock

PR #696, opened 2025-01-06
Update GetBySymfonyStringToConstructorInjectionRector fixture test to cover rector-srv:tv-readonly-add-ctor

samsonasik:
Ref https://github.com/rectorphp/rector-src/pull/6652


TomasVotruba:
Thanks :pray:

PR #700, opened 2025-02-14
Do not change method if name does not need to change

carlos-granados:
The `ActionSuffixRemoverRector` did not check if the new name was the same as the previous name and always tried to change it, resulting in the rule being reapplied with no need for it. This PR checks if the name needs to change and only updates the method if that is the case


samsonasik:
Thank you @carlos-granados

PR #701, opened 2025-02-15
[code-quality] Add InlineClassRoutePrefixRector

TomasVotruba:


trsteel88:
@TomasVotruba this rule isn't carrying the "name" from the prefix. This is resulting in broken route names.

In the example below, there is a route called "app_core.admin.organisation.index" prior to the rule running. After the rule runs, it becomes "index".

For example:

```php
#[Route(path: '/organisation', name: 'app_core.admin.organisation.')]
class OrganisationController extends AbstractController
{
    #[Route(name: 'index')]
    public function indexAction(): Response
    {
        ...
    }
```

Is being modified to:

```php
class OrganisationController extends AbstractController
{
    #[Route(path: '/organisation', name: 'index')]
    public function indexAction(): Response
    {
        ...
    }
```

But it should become:

```php
class OrganisationController extends AbstractController
{
    #[Route(path: '/organisation', name: 'app_core.admin.organisation.index')]
    public function indexAction(): Response
    {
        ...
    }
```

trsteel88:
Unfortunately I'm a bit swamped so I won't be able to. I'm not sure if it's been considered either but requirements, method etc and everything else will also need to be merged

TomasVotruba:
No wories :+1: 

cc @samsonasik  Could you look into fix for this one?

samsonasik:
I will try

samsonasik:
@trsteel88 hm.., I can't reproduce it, I probably missing something or missing some dependency, could you create simple reproducible repository for it? without it, i am not sure I can try provide a fix for it, thank you.

samsonasik:
@trsteel88 ok, I can reproduce on annotation usage, it seems concat string is needed if name is available.

samsonasik:
I created PR for it :

- https://github.com/rectorphp/rector-symfony/pull/704

PR #705, opened 2025-03-08
feat: add flysystem key

dsdeboer:
This PR adds the ability to auto convert the flysystem configuration

### DIFF
--- a/rules/CodeQuality/Rector/Closure/StringExtensionToConfigBuilderRector.php
+++ b/rules/CodeQuality/Rector/Closure/StringExtensionToConfigBuilderRector.php
@@ -51,6 +51,7 @@ final class StringExtensionToConfigBuilderRector extends AbstractRector
         'maker' => 'Symfony\Config\MakerConfig',
         'nelmio_cors' => 'Symfony\Config\NelmioCorsConfig',
         'api_platform' => 'Symfony\Config\ApiPlatformConfig',
+        // @see https://github.com/thephpleague/flysystem-bundle/blob/3.x/src/DependencyInjection/Configuration.php
         'flysystem' => 'Symfony\Config\FlysystemConfig',
     ];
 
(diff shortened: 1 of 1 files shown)


rules/CodeQuality/Rector/Closure/StringExtensionToConfigBuilderRector.php
```diff
@@ -51,6 +51,7 @@ final class StringExtensionToConfigBuilderRector extends AbstractRector
         'maker' => 'Symfony\Config\MakerConfig',
         'nelmio_cors' => 'Symfony\Config\NelmioCorsConfig',
         'api_platform' => 'Symfony\Config\ApiPlatformConfig',
+        'flysystem' => 'Symfony\Config\FlysystemConfig',
```
samsonasik, line 54: could you add `@see link` comment reference on it for future reference? Thank you.

dsdeboer: Ofcourse! Could you tell me where you want the link to point to?

samsonasik: just in above it, like:

```
       // @see ...
        'flysystem' => 'Symfony\Config\FlysystemConfig'

dsdeboer: If I understand correctly, 4148bb73fd878b4b6274d7be0ad4e488440f57d7 is what you meant.

samsonasik: not sure I understand how the class work, is `Symfony\Config\FlysystemConfig` auto created on the fly? since your reference pointed to different class

dsdeboer: Yes that's correct. If you installed `flysystem` then the added class is automatically generated based on the reference. This works the same as currently for the other keys.

samsonasik:
Thank you @dsdeboer

Issue #709, opened 2025-03-24
Bug: rule `ServiceSetStringNameToClassNameRector` applied for config.php


Myks92:
I have config in path `src/Access/Rbac/config.php`:

```php
// src/Access/Rbac/config.php
<?php

declare(strict_types=1);

namespace App\Access\Rbac;

use App\Access\Rbac\Entity\Permission;
use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;

return static function (ContainerConfigurator $configurator): void {
    $configurator->extension('doctrine', [
        'dbal' => [
            'types' => [
                Permission\IdType::NAME => Permission\IdType::class,
                Permission\NameType::NAME => Permission\NameType::class,
                Permission\CodeType::NAME => Permission\CodeType::class,
            ],
        ],
        'orm' => [
            'mappings' => [
                'AccessRbac' => [
                    'is_bundle' => false,
                    'type' => 'attribute',
                    'dir' => __DIR__ . '/Entity',
                    'prefix' => 'App\Access\Rbac\Entity',
                    'alias' => 'AccessRbac',
                ],
            ],
        ],
    ]);
};

```
In the config has mappings for Doctrine ORM. String ` 'prefix' => 'App\Access\Rbac\Entity'` replaced on `'prefix' => 'app\access\rbac\entity'` the rule `ServiceSetStringNameToClassNameRector` 

<img width="475" alt="Image" src="https://github.com/user-attachments/assets/6a996f98-1556-42b4-8ba4-8f0095a27def" />.

I think it's not right.

TomasVotruba:
Thank you for your report!

We'll need an isolated failing demo link from: http://getrector.com/demo,
that way we can reproduce the bug.

Myks92:
Unfortunately, I was unable to solve this problem using try Rector. Unfortunately, I was unable to solve this problem using try Rector. But I fixed up with PR #710.

PR #710, opened 2025-03-24
Skip excluded tag in container

Myks92:
Since https://github.com/symfony/symfony/pull/46279, Symfony Dependency Injection add the "container.exclude" to auto-discovered classes that are excluded by manual config. I also have made a pull request https://github.com/symfony/symfony/pull/50452 which solved the problem with displaying services in the console. But Rector does not consider this in any way. This PR solves this problem and also closes the issue #709


TomasVotruba:
Looks good, thank you for addition :people_hugging:

PR #711, opened 2025-03-28
Fix cases where rule was applied even when no code changed

carlos-granados:
Fixes all cases found using the code in this PR https://github.com/rectorphp/rector-src/pull/6794

### DIFF
--- a/rules/Symfony44/Rector/ClassMethod/ConsoleExecuteReturnIntRector.php
+++ b/rules/Symfony44/Rector/ClassMethod/ConsoleExecuteReturnIntRector.php
@@ -90,6 +90,8 @@ CODE_SAMPLE
             return null;
         }
 
+        $this->hasChanged = false;
+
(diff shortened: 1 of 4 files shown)


rules/Symfony44/Rector/ClassMethod/ConsoleExecuteReturnIntRector.php
```diff
@@ -90,6 +90,8 @@ public function refactor(Node $node): ?Node
             return null;
         }
 
+        $this->hasChanged = false;
```
carlos-granados, line 93: This was a case of shared state where the results of a previous test would affect the next tests. By adding this statement we make sure that all tests run independently

samsonasik:
Could you check other rector-* as well? thank you.

samsonasik:
Let's give it a try, thank you @carlos-granados

Issue #718, opened 2025-04-22
[Command] setHelp to attribute


JohJohan:
I saw the update from https://symfony.com/blog/new-in-symfony-7-3-invokable-commands-and-input-attributes and see we already have support to move to invoke but i saw it removed the `configure` option all together see diff:
```diff
#[AsCommand(name: 'some_name')]
final class SomeCommand
{
    public function __invoke(#[\Symfony\Component\Console\Attribute\Command\Argument]
    string $argument, #[\Symfony\Component\Console\Attribute\Command\Option]
    $option): int
    {
        $someArgument = $argument;
        $someOption = $option;

        // ...

        return 1;
    }

--    public function configure()
--    {
--        $this->setHelp('argument');
--        $this->addArgument('argument', InputArgument::REQUIRED, 'Argument description');
--        $this->addOption('option', 'o', InputOption::VALUE_NONE, 'Option description');
--    }
}
```
In this case we would want the `setHelp` we should move that one to the AsCommand see https://github.com/symfony/symfony/pull/59473 how should we handle this?

TomasVotruba:
Thanks for the report. Rule is in very early stages, based on couple diffs of referenced PR, so any test fixture of the right setup will help.

Can you send a PR with the test fixture you expecte?

JohJohan:
See: https://github.com/rectorphp/rector-symfony/pull/722

JohJohan:
Resolved with #722

PR #722, opened 2025-04-23
[Symfony 7.3] add constant and test for 7.3 and test help to attribute #718

JohJohan:
From: #718 
> Thanks for the report. Rule is in very early stages, based on couple diffs of referenced PR, so any test fixture of the right setup will help.
> 
> Can you send a PR with the test fixture you expect?

We would need:

1. Set support for symfony/console to 7.3 or at least define `AsCommand` has attribute `help`
2. Write a rector rule who will move help to the attribute (and make sure this is done before `InvokableCommandRector` as that will remove `configure` in total.


TomasVotruba:
Awesome, thanks for the test fixture :+1: 

Could you add fix as well?

rules-tests/Symfony73/Rector/Class_/InvokableCommandRector/Fixture/some_command.php.inc
```diff
@@ -45,6 +46,11 @@ use Symfony\Component\Console\Input\InputOption;
 #[AsCommand(name: 'some_name')]
 final class SomeCommand
 {
+    public function configure()
```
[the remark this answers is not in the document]
TomasVotruba: I think current state is better.

rules-tests/Symfony73/Rector/Class_/InvokableCommandRector/Fixture/some_command.php.inc
```diff
@@ -14,6 +14,7 @@ final class SomeCommand extends Command
 {
     public function configure()
     {
+        $this->setHelp('argument');
```
TomasVotruba, line 17: Could you add a new test fixture file only with `setHelp()` issue? 
That way we keep test fixtures separated by case. This makes them easier to maintain/debug later on

JohJohan: I added `some_command_with_set_help.php.inc` where its the same as here but then explicitly name this test as we need to have `$this->addArgumen()` and   `$this->addOption()` for the `InvokableCommandRector` to remove the `configure` bug like before

rules-tests/Symfony73/Rector/Class_/InvokableCommandRector/Fixture/some_command.php.inc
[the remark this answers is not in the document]
JohJohan: Okay then we can resolve this

TomasVotruba:
Looks good 👍

rules/Symfony73/Rector/Class_/CommandHelpToAttributeRector.php
```diff
@@ -0,0 +1,207 @@
...
+            return null;
+        }
+
+        $helpExpr = $this->findAndRemoveSetHelpExpr($configureClassMethod);
+        if (! $helpExpr instanceof String_) {
+            return null;
+        }
+
+        $wrappedHelp = new String_(
+            '<<<TXT ' . $helpExpr->value . ' TXT'
```
samsonasik, line 125: imo better use new line to ensure more support language, should be something like;

```suggestion
            '<<<TXT\n' . $helpExpr->value . '\nTXT'
```

JohJohan: Good point, i see Symfony also has newlines in there examples i added this and updated expected tests

tests/Set/Symfony73/Fixture/command_remove_config.php.inc
```diff
@@ -0,0 +1,68 @@
...
+namespace Rector\Symfony\Tests\Set\Symfony73\Fixture;
+
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputOption;
+
+#[AsCommand(name: 'some_name', help: '<<<TXT\nThe <info>%command.name%</info> command will do some
```
samsonasik, line 49: hm.., `\n` should be converted to new line, so the code maybe need in double quote so it applied as new line

rules/Symfony73/Rector/Class_/CommandHelpToAttributeRector.php
JohJohan: Your right double quote fixed that! I updated the tests again also updated the `CODE_SAMPLE`

tests/Set/Symfony73/Fixture/command_remove_config.php.inc
JohJohan: Your right double quote fixed that! I updated the tests again also updated the `CODE_SAMPLE`

tests/Set/Symfony73/Fixture/command_remove_config.php.inc
```diff
@@ -0,0 +1,70 @@
...
+namespace Rector\Symfony\Tests\Set\Symfony73\Fixture;
+
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputOption;
+
+#[AsCommand(name: 'some_name', help: '<<<TXT
```
samsonasik, line 49: no first single quote before TXT, but inside TXT

```suggestion
#[AsCommand(name: 'some_name', help: <<<'TXT'
```

tests/Set/Symfony73/Fixture/command_remove_config.php.inc
```diff
@@ -0,0 +1,70 @@
...
+The <info>%command.name%</info> command will do some
+        argument passed to it:
+
+        <info>php %command.full_name% argument</info>
+TXT')]
```
samsonasik, line 54: new line needs after TXT without "'" end quote

```suggestion
TXT
)]
```

rules/Symfony73/Rector/Class_/CommandHelpToAttributeRector.php
```diff
@@ -0,0 +1,209 @@
...
+            return null;
+        }
+
+        $helpExpr = $this->findAndRemoveSetHelpExpr($configureClassMethod);
+        if (! $helpExpr instanceof String_) {
+            return null;
+        }
+
+        $wrappedHelp = new String_(
+            "<<<TXT\n" . $helpExpr->value . "\nTXT"
```
samsonasik, line 127: I think you need to use  attribute KIND NOWDOC,  see

https://github.com/nikic/PHP-Parser/blob/c9d0b6c9f2c2c323661d31f7c81a26bf479ae3ff/lib/PhpParser/Node/Scalar/String_.php#L13

something like;

```php
$wrappedHelp = new String_( $helpExpr->value, [Attributekey::KIND => String_::KIND_NOWDOC]) 
```

tests/Set/Symfony73/Fixture/command_remove_config.php.inc
JohJohan: I updated my code to implement this and updated the tests

rules/Symfony73/Rector/Class_/CommandHelpToAttributeRector.php
JohJohan: I updated my code to build the `String()` like this but i got `TXT)]` in the end as there is no support in adding the new line after `TXT` is that okay? I think that is also a formatting thing and rector doesn't do that right

samsonasik: symfony 7.3 require php 8.2 so I think it is fine 👍

rules/Symfony73/Rector/Class_/CommandHelpToAttributeRector.php
```diff
@@ -0,0 +1,214 @@
...
+            function (Node $node) use (&$helpString): ?Expr {
+                if (! $node instanceof MethodCall) {
+                    return null;
+                }
+
+                if (! $this->isName($node->name, 'setHelp')) {
+                    return null;
+                }
+
+                $argExpr = $node->getArgs()[0]
```
samsonasik, line 178: ensure `[0]` index exists and not `firstClassCallable()` something like this pseudo code

```php
if ($node->isFirstClassCallable() || ! isset($node->getArgs()[0])) {
   return null;
}
```

rules/Symfony73/Rector/Class_/CommandHelpToAttributeRector.php
```diff
@@ -0,0 +1,214 @@
...
+     * Returns the argument passed to setHelp() and removes the MethodCall node.
+     */
+    private function findAndRemoveSetHelpExpr(ClassMethod $configureMethod): ?String_
+    {
+        $helpString = null;
+
+        $this->traverseNodesWithCallable(
+            (array) $configureMethod->stmts,
+            function (Node $node) use (&$helpString): ?Expr {
+                if (! $node instanceof MethodCall) {
```
samsonasik, line 170: don't change under inner `Class_` or `Function_`:

```php
if ($node instanceof \PhpParser\Node\Stmt\Class_ || $node instanceof \PhpParser\Node\Stmt\Function_) {
    return \PhpParser\NodeVisitor::DONT_TRAVERSE_CURRENT_AND_CHILDREN;
}
```

JohJohan: Added this if check

rules/Symfony73/Rector/Class_/CommandHelpToAttributeRector.php
```diff
@@ -0,0 +1,214 @@
...
+
+        if (! $this->isObjectType($node, new ObjectType(SymfonyClass::COMMAND))) {
+            return null;
+        }
+
+        $asCommandAttribute = $this->ensureAsCommandAttribute($node);
+
+        foreach ($asCommandAttribute->args as $arg) {
+            if ($arg->name?->toString() === 'help') {
+                return $node;
```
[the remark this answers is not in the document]
JohJohan: I made it so `CommandHelpToAttributeRector.php` wouldn't add `AsCommand` attribute when it doesn't exists i think its not the job of the rector rule it should have been done by `CommandConfigureToAttributeRector`

JohJohan:
@samsonasik note i did:

1. Added fixture to call all for `CommandConfigureToAttributeRector` (to check that the output will work for `CommandHelpToAttributeRector` and `InvokableCommandRector`
2. For `InvokableCommandRector` i made sure we also cleanup method chaining calls (also a fixture for this)
3. For `CommandConfigureToAttributeRector` added a fixture that has all attributes set from example `CommandConfigureToAttributeRector` rule and will add `help` attribute

TomasVotruba:
Thakn you @JohJohan  :clap:

PR #726, opened 2025-05-04
[config] Split off set to set provider for Symfony 3.0

TomasVotruba:


TomasVotruba:
Ref https://github.com/rectorphp/rector/issues/9107

PR #738, opened 2025-05-04
[symfony 5.3] split of configs per package

samsonasik:
Ref https://github.com/rectorphp/rector/issues/9107


samsonasik:

PR #739, opened 2025-05-05
[symfony 5.4] split of configs per package

samsonasik:


samsonasik:

config/sets/symfony/symfony5/symfony54/symfony54-security.php
```diff
@@ -0,0 +1,42 @@
+<?php
```
samsonasik, line 1: note to myself: extract to symfony54-security-core and symfony54-security-http

samsonasik: https://github.com/rectorphp/rector-symfony/pull/741

PR #744, opened 2025-05-06
[symfony 61] split of configs per package

samsonasik:


samsonasik:

PR #745, opened 2025-05-06
[symfony 62] split of configs per package

samsonasik:


config/sets/symfony/symfony6/symfony62.php
```diff
@@ -42,78 +19,15 @@
         new AnnotationToAttribute('Sensio\Bundle\FrameworkExtraBundle\Configuration\Template'),
```
samsonasik, line 19: The Sensio attribute require separate PR for re-take a look as not a symfony component.

Issue #760, opened 2025-05-08
[symfony 7.3] Add twig function attribute


TomasVotruba:
Ref https://symfony.com/blog/new-in-symfony-7-3-twig-extension-attributes

```php
namespace App\Twig;

use Twig\Attribute\AsTwigFilter;

class AppExtension
{
    #[AsTwigFilter('product_number')]
    public function formatProductNumber(string $number): string
    {
        // ...
    }
}
```

PR #761, opened 2025-05-09
[symfony 7.3] Add GetFiltersToAsTwigFilterAttributeRector to migrate #[TwigFilter]

TomasVotruba:

PR #764, opened 2025-05-09
kick off GetFunctionsToAsTwigFunctionAttributeRector

TomasVotruba:
Follow up to https://github.com/rectorphp/rector-symfony/pull/761

Ref https://symfony.com/blog/new-in-symfony-7-3-twig-extension-attributes

Covering https://github.com/symfony/symfony/pull/52748

Closes https://github.com/rectorphp/rector-symfony/issues/760

### DIFF
--- a/rules/Symfony73/NodeAnalyzer/LocalArrayMethodCallableMatcher.php
+++ b/rules/Symfony73/NodeAnalyzer/LocalArrayMethodCallableMatcher.php
@@ -1,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Symfony\Symfony73\NodeAnalyzer;
+
+use PhpParser\Node\Expr;
+use PhpParser\Node\Expr\Array_;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Scalar\String_;
+use Webmozart\Assert\Assert;
+
+final class LocalArrayMethodCallableMatcher
+{
+    public function match(Expr $expr): ?string
+    {
+        if ($expr instanceof MethodCall) {
+            Assert::isInstanceOf($expr->name, Identifier::class);
+
+            return $expr->name->toString();
+        }
+
+        if ($expr instanceof Array_) {
+            $firstItem = $expr->items[0];
+            if (! $firstItem->value instanceof Variable) {
+                return null;
+            }
+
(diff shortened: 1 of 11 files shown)


rules/Symfony73/NodeAnalyzer/LocalArrayMethodCallableMatcher.php
```diff
@@ -0,0 +1,50 @@
...
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Scalar\String_;
+use Webmozart\Assert\Assert;
+
+final class LocalArrayMethodCallableMatcher
+{
+    public function match(Expr $expr): ?string
+    {
+        if ($expr instanceof MethodCall) {
+            Assert::isInstanceOf($expr->name, Identifier::class);
```
samsonasik, line 20: MethodCall name can be any Expr, eg variable: `$foo->$bar()` so this assert can cause error, should if instanceof Identifier instanceof instead

TomasVotruba: Will update :+1:

TomasVotruba: Here we go 
https://github.com/rectorphp/rector-symfony/pull/766

rules/Symfony73/NodeAnalyzer/LocalArrayMethodCallableMatcher.php
```diff
@@ -0,0 +1,50 @@
...
+            if (! $firstItem->value instanceof Variable) {
+                return null;
+            }
+
+            $methodName = $firstItem->value->name;
+            if (! is_string($methodName)) {
+                return null;
+            }
+
+            if ($methodName !== 'this') {
```
samsonasik, line 36: Should be `$varObjectName` or something similar to avoid confusion of method name named "this"

TomasVotruba: Could you resolve these? I'm already in another rule

samsonasik: I will try

samsonasik: I created PR : 

- https://github.com/rectorphp/rector-symfony/pull/767

for it.

Issue #770, opened 2025-05-13
InvokableCommandInputAttributeRector generates invalid code.


mickverm:
Using:

- `rector/rector` 2.0.16
- `symfony/console` v7.3.0-BETA2

I started with the following command:

```php
<?php

declare(strict_types=1);

namespace App\Command;

use Symfony\Component\Console\Attribute\AsCommand;
use Symfony\Component\Console\Command\Command;
use Symfony\Component\Console\Input\InputArgument;
use Symfony\Component\Console\Input\InputInterface;
use Symfony\Component\Console\Output\OutputInterface;

#[AsCommand(
    name: 'app:hello',
    description: 'Greet the user.',
)]
class HelloCommand extends Command
{
    private const string ARGUMENT_NAME = 'name';

    #[\Override]
    protected function configure(): void
    {
        $this->addArgument(self::ARGUMENT_NAME, InputArgument::OPTIONAL, 'The name of the person to greet.');
    }

    #[\Override]
    protected function execute(InputInterface $input, OutputInterface $output): int
    {
        $name = $input->getArgument(self::ARGUMENT_NAME);
        $output->writeln("Hello {$name}!");

        return Command::SUCCESS;
    }
}
```

Which returns the following error:

```
[ERROR] Could not process "src\Command\HelloCommand.php" file, due
#0 vendor\rector\rector\vendor\rector\rector-symfony\rules\Symfony73\Rector\Class_\InvokableCommandInputAttributeRector.php(144): Rector\Symfony\Symfony73\NodeAnalyzer\CommandArgumentsAndOptionsResolver->collectCommandArguments()
#1 vendor\rector\rector\src\Rector\AbstractRector.php(112): Rector\Symfony\Symfony73\Rector\Class_\InvokableCommandInputAttributeRector->refactor()
#2 vendor\rector\rector\vendor\nikic\php-parser\lib\PhpParser\NodeTraverser.php(173): Rector\Rector\AbstractRector->enterNode()
#3 vendor\rector\rector\vendor\nikic\php-parser\lib\PhpParser\NodeTraverser.php(90): PhpParser\NodeTraverser->traverseArray()
#4 vendor\rector\rector\vendor\nikic\php-parser\lib\PhpParser\NodeTraverser.php(200): PhpParser\NodeTraverser->traverseNode()
[...]

PR #774, opened 2025-05-13
[Symfony 7.3] Keep $output param on InvokableCommandInputAttributeRector

samsonasik:
Ref https://github.com/rectorphp/rector-symfony/issues/770


samsonasik:
Fixed 🎉 /cc @mickverm

PR #776, opened 2025-05-14
[Symfony 7.3] Remove Argument argument: mode on InvokableCommandInputAttributeRector

mickverm:
Ref https://github.com/rectorphp/rector-symfony/issues/770


samsonasik:
Thank you @mickverm

PR #777, opened 2025-05-14
[Symfony 7.3] handle name from constant on InvokableCommandInputAttributeRector

samsonasik:
Ref https://github.com/rectorphp/rector-symfony/issues/770

PR #778, opened 2025-05-14
Fix ClassConstFetch use on InvokableCommandInputAttributeRector

samsonasik:
Ref https://github.com/rectorphp/rector-symfony/pull/777#pullrequestreview-2839570919


rules-tests/Symfony73/Rector/Class_/InvokableCommandInputAttributeRector/Fixture/name_from_constant.php.inc
```diff
@@ -55,7 +55,7 @@ class NameFromConstant
     protected function __invoke(#[\Symfony\Component\Console\Attribute\Argument(name: self::ARGUMENT_NAME, description: 'The name of the person to greet.')]
     ?string $name, OutputInterface $output): int
     {
-        $name = self::ARGUMENT_NAME;
+        $name = $name;
```
samsonasik, line 58: @mickverm this `$name = $name` equal is OK, it can be removed via `Rector\DeadCode\Rector\Expression\SimplifyMirrorAssignRector` with activate `deadCode` set

https://getrector.com/demo/29ca0dfe-2bf8-4ff9-ac6f-8f1b5e5161b3

PR #779, opened 2025-05-14
[Symfony 7.3] Make __invoke() public in InvokableCommandInputAttributeRector

mickverm:
```
Warning: The magic method App\Command\HelloCommand::__invoke() must have public visibility
```

Ref #770

### DIFF
--- a/rules/Symfony73/Rector/Class_/InvokableCommandInputAttributeRector.php
+++ b/rules/Symfony73/Rector/Class_/InvokableCommandInputAttributeRector.php
@@ -19,6 +19,7 @@ use PhpParser\Node\Stmt\Expression;
 use Rector\Doctrine\NodeAnalyzer\AttributeFinder;
 use Rector\Exception\ShouldNotHappenException;
 use Rector\PhpParser\Node\Value\ValueResolver;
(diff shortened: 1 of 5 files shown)


rules/Symfony73/Rector/Class_/InvokableCommandInputAttributeRector.php
```diff
@@ -147,6 +148,7 @@ public function refactor(Node $node): ?Class_
         }
 
         $executeClassMethod->name = new Identifier('__invoke');
+        $executeClassMethod->flags = Modifiers::PUBLIC;
```
samsonasik, line 151: use `\Rector\Privatization\NodeManipulator\VisibilityManipulator::makePublic()` instead to avoid invalid replace in case of there is another flag combination.

https://github.com/rectorphp/rector-src/blob/8dfc7aa424967b1f8e509e125706f19c0cf9cbb6/rules/Privatization/NodeManipulator/VisibilityManipulator.php#L92-L95

mickverm: Done!

PR #792, opened 2025-06-05
Fix namespace of console Option attribute

gharlan:


samsonasik:
verified

https://github.com/symfony/symfony/blob/v7.3.0/src/Symfony/Component/Console/Attribute/Option.php

thank you @gharlan

PR #795, opened 2025-06-05
[Symfony 7.3] Handle fluent call addArgument and addOption on InvokableCommandInputAttributeRector

samsonasik:
Closes https://github.com/rectorphp/rector-symfony/pull/794


samsonasik:
Fixed :tada: /cc @gharlan

PR #805, opened 2025-06-30
Twig attributes: make methods public

gharlan:
My twig function/filter methods are often private. But when using the new attributes, they have to be public.


TomasVotruba:
I see. Thanks for the description and the fix @gharlan :+1:

PR #809, opened 2025-07-08
[Symfony 74] Rename Application::add() to Application::addCommand()

mttsch:


mttsch:
Not sure if you already want to merge already it with Symfony 7.4 not being released yet. If not, the PR can be kept open.

samsonasik:
Let's merge, we can revisit/update when symfony 7.4 released if there is updated method name.

PR #811, opened 2025-07-10
InvokableCommandInputAttributeRector test fixture

darthf1:
fixture for: https://github.com/rectorphp/rector/issues/9254

### DIFF
--- a/rules-tests/Symfony73/Rector/Class_/InvokableCommandInputAttributeRector/Fixture/command_with_io.php.inc
+++ b/rules-tests/Symfony73/Rector/Class_/InvokableCommandInputAttributeRector/Fixture/command_with_io.php.inc
@@ -1,0 +1,42 @@
+<?php
+
+namespace Rector\Tests\Symfony\Symfony73\Rector\Class_\InvokableCommandInputAttributeRector\Fixture;
+
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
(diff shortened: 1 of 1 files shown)


rules-tests/Symfony73/Rector/Class_/InvokableCommandInputAttributeRector/Fixture/command_with_io.php.inc
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Rector\Tests\Symfony\Symfony73\Rector\Class_\InvokableCommandInputAttributeRector\Fixture;
+
+use Symfony\Component\Console\Attribute\AsCommand;
+ 
+#[AsCommand('app:my-command')]
+final class MyCommand
+{
+    #[\Override]
+    public function __invoke(\Symfony\Component\Console\Style\SymfonyStyle $io): int
```
samsonasik, line 35: I think it should not be replaced, skip the change should be the way, right? /cc @darthf1 @thePanz @gharlan could you verify? Thank you.

thePanz: as this is a bugfix, I would keep a KISS solution: @samsonasik 's suggestion to not rewrite the parameters in the constructor looks reasonable as a first step.

TomasVotruba:
@darthf1 I've cherry picked your commit in https://github.com/rectorphp/rector-symfony/pull/838
Thanks :+1:

PR #816, opened 2025-07-15
[SYMFONY 6.3] replace Link argument values with constants

JohJohan:
See https://github.com/rectorphp/rector-symfony/pull/812 & https://github.com/rectorphp/rector/issues/9256

On hold see: https://github.com/rectorphp/rector-src/pull/7068


samsonasik:
Thank you @JohJohan

PR #817, opened 2025-07-15
[SYMFONY 3.2] replace Cookie argument values with constants

JohJohan:
See: https://github.com/rectorphp/rector/issues/9256

On hold see: https://github.com/rectorphp/rector-src/pull/7068


samsonasik:
Thank you @JohJohan

PR #823, opened 2025-07-17
Update Fixtures for GetFunctionsToAsTwigFunctionAttributeRector

cvergne:
As requested in rectorphp/rector#9361, here is a draft Pull Request with :
- New fixture for the case with a static method from an other class in the file
- Updated fixtures for the valid case to remove the extends of `AbstractExtension`

### DIFF
--- a/rules-tests/Symfony73/Rector/Class_/GetFunctionsToAsTwigFunctionAttributeRector/Fixture/skip_difference_class.php.inc
+++ b/rules-tests/Symfony73/Rector/Class_/GetFunctionsToAsTwigFunctionAttributeRector/Fixture/skip_difference_class.php.inc
@@ -1,0 +1,23 @@
+<?php
+
+namespace Rector\Symfony\Tests\Symfony73\Rector\Class_\GetFunctionsToAsTwigFunctionAttributeRector\Fixture;
+
+use Twig\Extension\AbstractExtension;
+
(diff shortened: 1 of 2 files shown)


rules-tests/Symfony73/Rector/Class_/GetFunctionsToAsTwigFunctionAttributeRector/Fixture/skip_difference_class.php.inc
```diff
@@ -0,0 +1,23 @@
...
+namespace Rector\Symfony\Tests\Symfony73\Rector\Class_\GetFunctionsToAsTwigFunctionAttributeRector\Fixture;
+
+use Twig\Extension\AbstractExtension;
+
+final class SkipDifferentClass extends AbstractExtension
+{
+    public function getFunctions(): array
+    {
+        return [
+            new \Twig\TwigFunction('some_function', [DifferentClass, 'someFunction']),
```
samsonasik, line 12: this should be `DifferentClass::class` ?

```suggestion
            new \Twig\TwigFunction('some_function', [DifferentClass::class, 'someFunction']),
```

cvergne: Totally, sorry, I've fixed it

PR #824, opened 2025-07-23
Keep AsCommand in InvokableCommandInputAttributeRector documentation

bobvandevijver:
This rule does not touch the `AsCommand` attribute at all, as it shouldn't because otherwise it would no longer be a command!

See also for examples this [test](https://github.com/rectorphp/rector-symfony/blob/main/rules-tests/Symfony73/Rector/Class_/InvokableCommandInputAttributeRector/Fixture/some_command.php.inc).


samsonasik:
Thank you @bobvandevijver

PR #827, opened 2025-08-12
[CodeQuality] Skip got duplicated method after suffix removed on ActionSuffixRemoverRector

samsonasik:
Cause can't redeclare method error.


samsonasik:
Fixed 🎉

samsonasik:

PR #906, opened 2025-12-16
DRAFT feat: flag symfony controller when using AsController attribute, make allowed func args configurable

tony-stark-eth:
Not that much time currently but still felt like a valuable contribution, feel free to complete or close if not desired.

Flagging a Class as Controller when using Route attributes would also be a nice addition.


samsonasik (changes requested):
this looks like multiple functionalities, I think you can extract into 2 PRs:

- check controller via AsController attribute
- allow configurable known objects

add both PRs with the fixture to apply/skip.

TomasVotruba:
Closing for lack of feedback last 6 months. If this is still active, please reopen passing PR :+1:

PR #913, opened 2026-01-13
[Symfony 7.3] Replace AuthorizationChecker with AccessDecisionManager…

meddrh:
Adds a Symfony 7.3 Rector rule that replaces `AuthorizationCheckerInterface`
with `AccessDecisionManagerInterface` in voters.

The rule updates injected services, renames properties, and converts
`isGranted()` calls to `decide()`.

Tests and fixtures are included.

### DIFF
--- a/rules/Symfony73/Rector/Class_/AuthorizationCheckerToAccessDecisionManagerInVoterRector.php
+++ b/rules/Symfony73/Rector/Class_/AuthorizationCheckerToAccessDecisionManagerInVoterRector.php
@@ -1,0 +1,217 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Symfony\Symfony73\Rector\Class_;
+
+use PhpParser\Node;
+use PhpParser\Node\Arg;
+use PhpParser\Node\ArrayItem;
+use PhpParser\Node\Expr\Array_;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\PropertyFetch;
+use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Name\FullyQualified;
+use PhpParser\Node\Stmt\Class_;
+use PhpParser\Node\Stmt\ClassMethod;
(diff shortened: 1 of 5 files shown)


khadidja34 (approved):

rules/Symfony73/Rector/Class_/AuthorizationCheckerToAccessDecisionManagerInVoterRector.php
```diff
@@ -0,0 +1,211 @@
...
+                $hasChanged = true;
+            }
+        }
+
+        // 3) Replace isGranted() with decide()
+        $voteMethod = $node->getMethod('voteOnAttribute');
+        if ($voteMethod instanceof ClassMethod) {
+            $this->traverseNodesWithCallable(
+                $voteMethod,
+                function (Node $node) use (&$hasChanged, $voteMethod, $renamedProperties) {
```
samsonasik: add to not traverse inner class and function:

```php
if ($node instanceof Class_ || $node instanceof Function_) {
     return NodeVisitor::DONT_TRAVERSE_CURRENT_AND_CHILDREN;
}
```

samsonasik (changes requested):
Please register the rule to

https://github.com/rectorphp/rector-symfony/blob/main/config/sets/symfony/symfony7/symfony73/symfony73-security-core.php

samsonasik (approved):

trsteel88:
@samsonasik - is is ruining my Voters without disabling it. I feel like this shouldn't be default.

I have a cascading hierarchy of voters and I need to inject the AuthorizationCheckerInterface so i can cascade the validation up the tree

samsonasik:
@trsteel88 I think you can skip the rule https://getrector.com/documentation/ignoring-rules-or-paths

trsteel88:
@samsonasik Yeh I have but not sure why this is a default rule? Injecting the AuthorisationCheckerInterface into a Voter is a valid use-case

samsonasik:
@trsteel88 could you create failing test case where the transform should be skipped? thank you.

PR #929, opened 2026-03-20
[CodeQuality] Handle different parameter type same name used on ControllerMethodInjectionToConstructorRector

### LINKED ISSUE #9697: ControllerMethodInjectionToConstructorRector confused by parameters with the same name
# Bug Report

This seems pretty similar to <https://github.com/rectorphp/rector/issues/9688>, though the parameters are now being left alone - both correctly for method `three` and incorrectly for method `two` - (possibly due to [the changes made](https://github.com/rectorphp/rector-symfony/pull/925) towards addressing that issue). The bigger problem is that the usage is being replaced with the property, regardless.

See https://getrector.com/demo/ede48a24-90f9-4316-8b9a-b3a3554e5f8a

```php
<?php

class FooController extends AbstractController
{
    public function one(Foo $service)
    {
        dump($service);
    }

    public function two(Foo $service)
    {
        dump($service);
    }

    public function three(Bar $service)
    {
        dump($service);
    }
}
```

### Responsible rules

* `ControllerMethodInjectionToConstructorRector`

## Expected Behavior

Either:
```diff
<?php

use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;

class FooController extends AbstractController
{
+    public function __construct(private readonly Foo $service)
+    {
+    }
+
-    public function one(Foo $service)
+    public function one()
    {
-        dump($service);
+        dump($this->service);
    }
 
-    public function two(Foo $service)
+    public function two()
    {
-        dump($service);
+        dump($this->service);
    }
 
    public function three(Bar $service)
    {
        dump($service);
    }
}

```
(or skipping the parameters with the same name).

Thanks!

samsonasik:
Closes https://github.com/rectorphp/rector/issues/9697

### DIFF
--- a/rules-tests/CodeQuality/Rector/Class_/ControllerMethodInjectionToConstructorRector/Fixture/different_parameter_type_same_name_used.php.inc
+++ b/rules-tests/CodeQuality/Rector/Class_/ControllerMethodInjectionToConstructorRector/Fixture/different_parameter_type_same_name_used.php.inc
@@ -1,0 +1,58 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Symfony\Tests\CodeQuality\Rector\Class_\ControllerMethodInjectionToConstructorRector\Fixture;
+
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Looks good :+1:

PR #931, opened 2026-04-14
Fix unit test due to new PHPStan release

samsonasik:
Ref error https://github.com/rectorphp/rector-src/actions/runs/24381475596/job/71205726476?pr=7969#step:7:30

https://github.com/phpstan/phpstan/releases/tag/2.1.47


samsonasik:
All green 🎉 @TomasVotruba I am merging it ;)

PR #934, opened 2026-05-16
fix ControllerMethodInjectionToConstructorRector - update call sites when params are moved to constructor

### LINKED ISSUE #9695: ControllerMethodInjectionToConstructorRector removing parameters from definitions but not calls
# Bug Report

See https://getrector.com/demo/44eaf6a6-8742-4d3e-a15b-be0127df8d83

```php
<?php

class ExampleController extends AbstractController
{
    public function example(Foo $foo)
    {
        $this->problematic($foo);
    }
    
    public function problematic(Foo $foo, ?string $optional = "test")
    {
        dump($foo);
        dump($optional);
    }
}
```

### Responsible rules

* `ControllerMethodInjectionToConstructorRector`

## Expected Behaviour

```diff


 <?php
 
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;

 
 class ExampleController extends AbstractController
 {
+    public function __construct(private readonly \Foo $foo)
+    {
+    }
+
-    public function example(Foo $foo)
+    public function example()
     {
-        $this->problematic($foo);
+        $this->problematic();
     }

-    public function problematic(Foo $foo, ?string $optional = "test")
+    public function problematic(?string $optional = "test")
     {
-        dump($foo);
+        dump($this->foo);
         dump($optional);
     }
 }
```

Thanks!

TomasVotruba:
## Problem

Fixes https://github.com/rectorphp/rector/issues/9695

When a controller has a helper method that takes both a service parameter **and** optional scalar parameters:

```php
final class SomeController extends AbstractController
{
    #[Route('/some-action')]
    public function someAction(LoggerInterface $logger)
    {
        $this->problematic($logger);
    }

    public function problematic(LoggerInterface $logger, ?string $optional = 'test')
    {
        $logger->log('level', 'value');
        dump($optional);
    }
}
```

The rule correctly moved `LoggerInterface $logger` out of both method signatures and into the constructor, and replaced `$logger` usages inside each method body with `$this->logger`. However, it did **not** update the **call site** `$this->problematic($logger)` in `someAction` — after `$logger` was replaced with `$this->logger` by the variable-replacement step, the call became `$this->problematic($this->logger)`, even though `problematic` no longer accepts that argument. This produced invalid code.

## Fix

After all param removals and variable replacements, a new `updateCallSitesForRemovedParams` step traverses every method body in the class. For any `$this->methodName(...)` call whose method had parameters removed, the corresponding positional arguments are removed from the call.

## Changes

- `rules/CodeQuality/Rector/Class_/ControllerMethodInjectionToConstructorRector.php` — track removed param positions per method; add `updateCallSitesForRemovedParams()` that strips the stale arguments from internal call sites
- `rules-tests/.../Fixture/internal_method_call_arg_removal.php.inc` — new fixture reproducing the exact scenario from the issue

---
_Generated by [Claude Code](https://claude.ai/code/session_01NNfB4GsBWmAqPas9n22r2H)_

### DIFF
--- a/rules-tests/CodeQuality/Rector/Class_/ControllerMethodInjectionToConstructorRector/Fixture/internal_method_call_arg_removal.php.inc
+++ b/rules-tests/CodeQuality/Rector/Class_/ControllerMethodInjectionToConstructorRector/Fixture/internal_method_call_arg_removal.php.inc
@@ -1,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Symfony\Tests\CodeQuality\Rector\Class_\ControllerMethodInjectionToConstructorRector\Fixture;
+
+use Psr\Log\LoggerInterface;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\Routing\Annotation\Route;
+
+final class InternalMethodCallArgRemoval extends AbstractController
+{
+    #[Route('/some-action', name: 'some_action')]
+    public function someAction(LoggerInterface $logger)
+    {
+        $this->problematic($logger);
+    }
+
+    public function problematic(LoggerInterface $logger, ?string $optional = 'test')
+    {
+        $logger->log('level', 'value');
+        dump($optional);
+    }
+}
+
+?>
+-----
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Good job :)

PR #938, opened 2026-05-24
[Symfony81] add rule for Ulid::isValid() new argument

MrYamous:
New rule to adress [Uid changes](https://github.com/symfony/symfony/blob/8.1/UPGRADE-8.1.md#uid)

> Add argument $format to Ulid::isValid()

### DIFF
--- a/rules/Symfony81/Rector/StaticCall/AddFormatArgumentToIsValidRector.php
+++ b/rules/Symfony81/Rector/StaticCall/AddFormatArgumentToIsValidRector.php
@@ -1,0 +1,94 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Symfony\Symfony81\Rector\StaticCall;
+
+use PhpParser\Node;
+use PhpParser\Node\Arg;
+use PhpParser\Node\Expr\ClassConstFetch;
+use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Name;
+use PhpParser\Node\Name\FullyQualified;
+use PhpParser\Node\Name\Relative;
+use PhpParser\Node\Stmt\Class_;
+use Rector\Rector\AbstractRector;
+use Rector\Symfony\Enum\SymfonyClass;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+/**
+ * @see https://github.com/symfony/symfony/blob/8.1/UPGRADE-8.1.md#uid
+ *
+ * @see \Rector\Symfony\Tests\Symfony81\Rector\StaticCall\AddFormatArgumentToIsValidRector\AddFormatArgumentToIsValidRectorTest
+ */
+final class AddFormatArgumentToIsValidRector extends AbstractRector
+{
+    public function getRuleDefinition(): RuleDefinition
+    {
+        return new RuleDefinition(
+            'Add $format argument to Ulid::isValid()',
+            [
+                new CodeSample(
+                    <<<'CODE_SAMPLE'
+                        Symfony\Component\Uid\Ulid;
+
+                        final class Foo
+                        {
+                            public function bar(string $id): bool
+                            {
+                                return Ulid::isValid($id);
+                            }
+                        }
+                        CODE_SAMPLE
+                    ,
+                    <<<'CODE_SAMPLE'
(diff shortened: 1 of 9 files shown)


TomasVotruba:
CI needs fixing. Thanks 🙏

rules/Symfony81/Rector/StaticCall/AddFormatArgumentToIsValidRector.php
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        if (! $this->isName($node->name, 'isValid')) {
+            return null;
+        }
+
+        if (count($node->args) === 2) {
+            return null;
+        }
+
```
samsonasik, line 87: this needs skip on first class callable:

```php
if ($node->isFirstClassCallable()) {
    return null;
}
```

rules/Symfony81/Rector/StaticCall/AddFormatArgumentToIsValidRector.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+        if (! $this->isName($node->class, 'Symfony\Component\HttpFoundation\BinaryFileResponse')) {
+            return null;
+        }
+
+        if (! $this->isName($node->name, 'isValid')) {
+            return null;
+        }
+
+        if (count($node->args) === 2) {
```
samsonasik: should use `!==1` instead ?

MrYamous: I had doubts while writing it, but I think you're right

rules/Symfony81/Rector/StaticCall/AddFormatArgumentToIsValidRector.php
```diff
@@ -0,0 +1,91 @@
...
+use Rector\Symfony\Enum\SymfonyClass;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+/**
+ * @see https://github.com/symfony/symfony/blob/8.1/UPGRADE-8.1.md#uid
+ *
+ * @see \Rector\Symfony\Tests\Symfony81\Rector\StaticCall\AddFormatArgumentToIsValidRector\AddFormatArgumentToIsValidRectorTest
+ */
+final class AddFormatArgumentToIsValidRector extends AbstractRector
```
samsonasik, line 25: register the rule to symfony 8.1 set, should be at https://github.com/rectorphp/rector-symfony/tree/main/config/sets/symfony/symfony8

samsonasik: also on https://github.com/rectorphp/rector-symfony/blob/main/src/Set/SetProvider/Symfony8SetProvider.php and https://github.com/rectorphp/rector-symfony/blob/main/src/Set/SymfonySetList.php

MrYamous: Thanks ! There is a lot of things to add 😅

rules/Symfony81/Rector/StaticCall/AddFormatArgumentToIsValidRector.php
```diff
@@ -0,0 +1,94 @@
...
+        if ($node->isFirstClassCallable()) {
+            return null;
+        }
+
+        if (\count($node->args) !==1) {
+            return null;
+        }
+
+        $node->args[] = new Arg(
+            new ClassConstFetch(new FullyQualified(SymfonyClass::ULID_CLASS), 'FORMAT_BASE_32')
```
MrYamous, line 89: I'm not sure about this. I think it make no sense to use FQCN here because `Ulid` class is used anyway, but I didn't find something better to use than `FullyQualified` or `Relative`

samsonasik: `FullyQualified` is fine 👍 as ensure it always works and pointed to correct class. 

For auto import, there is already `->withImportNames()` for it :)

MrYamous: Then it looks ready

samsonasik (approved):

MrYamous:
Rebase done

TomasVotruba:
Thank you :+1:  Let's ship it

PR #942, opened 2026-05-29
[Symfony81] Add new rule for Security component

MrYamous:
Add a new rule for [Security component](https://github.com/symfony/symfony/blob/8.2/UPGRADE-8.1.md#security)

> Deprecate passing the $eraseCredentials argument to AuthenticatorManager::__construct(), as the eraseCredentials() method was removed in Symfony 8.0

Code implementation is a bit more precise : 
>      * @param ExposeSecurityLevel                     $exposeSecurityErrors Passing a bool is deprecated since Symfony 8.1

[Codebase for reference](https://github.com/symfony/symfony/blob/179bc110d25737b7dc01df43c63f257b53cc7186/src/Symfony/Component/Security/Http/Authentication/AuthenticatorManager.php#L49-L66)

### DIFF
--- a/config/sets/symfony/symfony8/symfony81/symfony81-security.php
+++ b/config/sets/symfony/symfony8/symfony81/symfony81-security.php
@@ -1,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+use Rector\Config\RectorConfig;
+use Rector\Symfony\Symfony81\Rector\New_\RemoveEraseCredentialsFromAuthenticatorManagerRector;
+
+return static function (RectorConfig $rectorConfig): void {
+    $rectorConfig->rules([RemoveEraseCredentialsFromAuthenticatorManagerRector::class]);
+};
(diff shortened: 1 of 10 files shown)


TomasVotruba:
CI needs fixing

MrYamous:
> CI needs fixing

@TomasVotruba CI fixed

config/sets/symfony/symfony8/symfony81/symfony81-security.php
```diff
@@ -0,0 +1,10 @@
+<?php
```
samsonasik, line 1: this needs to be included in  `config/sets/symfony/symfony8/symfony81.php` as well

https://github.com/rectorphp/rector-symfony/blob/main/config/sets/symfony/symfony8/symfony81.php

TomasVotruba: I'll see if we can make PHPStan rule from this, to avoid missing any includes

TomasVotruba:
Thank you @MrYamous :+1:

config/sets/symfony/symfony8/symfony81/symfony81-security.php
TomasVotruba: Here we go :) https://github.com/rectorphp/rector-symfony/pull/943

MrYamous: Thanks ! I need a checklist on my side 😅

PR #899, opened 2025-12-09
Order of optional parameters

florisluiten:
Expected output and potential fix for issue with InvokableCommandInputAttributeRector

PR #944, opened 2026-05-30
Order optional command parameters last in InvokableCommandInputAttributeRector

### LINKED ISSUE #9546: Parameter order for optional parameters in InvokableCommandInputAttributeRector
# Bug Report

| Rector version | 2.2.7 |

Using InvokableCommandInputAttributeRector, it seems that optional and required parameters are mixed depending on where they appear in the code. This may lead to a parameterlist that has optional parameters before required parameters.

https://github.com/rectorphp/rector-symfony/pull/899

## Expected Behaviour

Imo, the optional parameters should always be listed last.

TomasVotruba:
Fixes rectorphp/rector#9546

## Problem

`InvokableCommandInputAttributeRector` emitted `__invoke()` parameters in the same order the `addArgument()`/`addOption()` calls appeared in `configure()`. When an optional parameter (one with a default value) was declared before a required one, the generated method ended up with an optional parameter in front of a required parameter — which is invalid PHP and would have to be fixed up afterwards by `OptionalParametersAfterRequiredRector`.

## Fix

After building the `__invoke()` parameter list, parameters are partitioned so required parameters (no default) keep their relative order and come first, followed by optional parameters (with a default). This guarantees optional parameters are always listed last.

This is a fresh, passing take on the previously-closed #899 — the trailing-comma formatting problem reported there no longer reproduces with the current printer.

## Tests

- Updated `option_with_array_type.php.inc` fixture to reflect the corrected order.
- Added `optional_option_before_required.php.inc` fixture covering an optional option declared before a required one.

### DIFF
--- a/rules-tests/Symfony73/Rector/Class_/InvokableCommandInputAttributeRector/Fixture/optional_option_before_required.php.inc
+++ b/rules-tests/Symfony73/Rector/Class_/InvokableCommandInputAttributeRector/Fixture/optional_option_before_required.php.inc
@@ -1,0 +1,62 @@
+<?php
+
+namespace Rector\Symfony\Tests\Symfony73\Rector\Class_\InvokableCommandInputAttributeRector\Fixture;
+
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Already fixed, just adding the fixture here

PR #945, opened 2026-05-30
[Symfony73] Sort optional parameters last in InvokableCommandInputAttributeRector

### LINKED ISSUE #9546: Parameter order for optional parameters in InvokableCommandInputAttributeRector

TomasVotruba:
Fixes rectorphp/rector#9546

`InvokableCommandInputAttributeRector` could emit `__invoke()` signatures where an optional parameter (one with a default value) appeared before a required one, producing an invalid PHP signature.

This sorts the generated parameters so all required parameters come first and optional ones last, while keeping their relative order otherwise. The `$output` parameter (no default) is treated as required and therefore stays among the leading params.

This is a fresh take on the approach from #899 (closed for inactivity); building the params array from scratch avoids the trailing-comma formatting glitches reported there.

Covered by:
- updated `Fixture/DefaultValue/option_with_array_type.php.inc` (previously asserted the buggy order)
- new `Fixture/with_default_and_without_default_value.php.inc`

### DIFF
--- a/rules-tests/Symfony73/Rector/Class_/InvokableCommandInputAttributeRector/Fixture/DefaultValue/option_with_array_type.php.inc
+++ b/rules-tests/Symfony73/Rector/Class_/InvokableCommandInputAttributeRector/Fixture/DefaultValue/option_with_array_type.php.inc
@@ -43,10 +43,10 @@ use Symfony\Component\Console\Output\OutputInterface;
 class OptionWithOptionalValue
 {
     public function __invoke(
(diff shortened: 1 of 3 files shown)


TomasVotruba:
LGTM :+1:

PR #947, opened 2026-06-02
fix: only convert Twig extensions fully reducible to AsTwig attributes

### LINKED ISSUE #9775: Incorrect behavior of GetFiltersToAsTwigFilterAttributeRector  and GetFunctionsToAsTwigFunctionAttributeRector
| Rector version | 2.4.5 |
| Rules | GetFiltersToAsTwigFilterAttributeRector and GetFunctionsToAsTwigFunctionAttributeRector |

Related to Symfony 7.3 sets.

Hello, 

These rules remove "extends AbstractExtension" from Twig extensions, even when they still need it.

In some cases, the file should not be processed, as "extends AbstractExtension" is still needed to be registered as a Twig extension, such as when:
- the file implements GlobalsInterface
- there remain some parent functions (ex: getTests(), ...)
- when the cleaning cannot be completed.
For example, the callable of a function or filter can be a callable outside of the class itself, or null.

Ex:
```
new TwigFunction('my_form_function', null, ['node_class' => RenderBlockNode::class, 'is_safe' => ['html']]),
```

```
new TwigFunction('function_name', [$this->myInjectedService, 'serviceFunction']),
```

In these cases, the file is still treated by the rule, but some "getFunctions" or "getFilters" declarations are left intact. However, as the "extends AbstractExtension" is removed (it should be, as attributes + extends are incompatible), the remaining functions or filters are never registered.

The question is, are these 2 rules relevant if they are separated? Because in a file where there are both getFilters and getFunctions, assuming they can be processed, the rules cannot be used independently.

Also, there is no rule for the "AsTwigTest" attribute.

So, due to the nature of the problem (lots of cases where the file should not be processed), should these rules exist?

I see 3 options:
- remove the rules as it's too complex to be certain the file is processable (is it?)
- merge the 2 rules, adding support for AsTwigTest.
- leave as is (but it leads to a 500 error).

Any opinions?

TomasVotruba:
Fixes rectorphp/rector#9775

## Problem

`GetFiltersToAsTwigFilterAttributeRector` and `GetFunctionsToAsTwigFunctionAttributeRector` removed `extends AbstractExtension` even when the class still needed it, leaving some filters/functions unregistered (and producing broken extensions).

This happened in several cases reported in the issue:

- A registration whose callable cannot become an attribute on a local method, e.g. a `null` callable or an injected-service callable:
  ```php
  new TwigFunction('my_form_function', null, ['is_safe' => ['html']]),
  new TwigFunction('function_name', [$this->myInjectedService, 'serviceFunction']),
  ```
  These items were silently left in `getFunctions()`/`getFilters()` while `extends AbstractExtension` was still dropped.
- The class still defines other Twig extension hooks (`getTests()`, `getGlobals()`, the *other* `get*()` method when only one of the two rules runs, …) or implements `GlobalsInterface`. The parent class is still required to register those, but it was removed.

## Fix

The shared `GetMethodToAsTwigAttributeTransformer` is now **all-or-nothing**:

1. Every item in the returned array is validated *before* any mutation. If a single registration cannot be fully expressed as an attribute on a local method, the class is left untouched (also fixes a latent partial-mutation bug where earlier items were already rewritten before a later item caused a bail-out).
2. `extends AbstractExtension` is only removed — and the conversion only applied — when the class no longer needs the parent extension, i.e. it has no other Twig registration methods (`getTokenParsers`, `getNodeVisitors`, `getFilters`, `getTests`, `getFunctions`, `getOperators`, `getGlobals`) and does not implement `GlobalsInterface`/`ExtensionInterface`.

This guarantees the rules never produce an extension where some filters/functions silently stop being registered.

## Tests

Added skip fixtures covering the reported scenarios:

- `null` callable mixed with a convertible one
- injected-service callable `[$this->service, 'method']`
- remaining `getFilters()` when converting `getFunctions()` (and `getTests()` when converting `getFilters()`)
- class implementing `GlobalsInterface`

All existing fixtures (full, single-method conversions) still convert as before.

### DIFF
--- a/rules-tests/Symfony73/Rector/Class_/GetFiltersToAsTwigFilterAttributeRector/Fixture/skip_remaining_get_tests.php.inc
+++ b/rules-tests/Symfony73/Rector/Class_/GetFiltersToAsTwigFilterAttributeRector/Fixture/skip_remaining_get_tests.php.inc
@@ -1,0 +1,32 @@
+<?php
+
+namespace Rector\Symfony\Tests\Symfony73\Rector\Class_\GetFiltersToAsTwigFilterAttributeRector\Fixture;
+
+use Twig\Extension\AbstractExtension;
+
+final class SkipRemainingGetTests extends AbstractExtension
+{
+    public function getFilters(): array
+    {
+        return [
+            new \Twig\TwigFilter('some_filter', [$this, 'someFilter']),
+        ];
+    }
+
+    public function getTests(): array
+    {
+        return [
+            new \Twig\TwigTest('some_test', [$this, 'someTest']),
+        ];
+    }
+
+    public function someFilter($value)
+    {
+        return $value;
+    }
+
+    public function someTest($value)
+    {
+        return $value;
+    }
+}
--- a/rules-tests/Symfony73/Rector/Class_/GetFunctionsToAsTwigFunctionAttributeRector/Fixture/skip_globals_interface.php.inc
+++ b/rules-tests/Symfony73/Rector/Class_/GetFunctionsToAsTwigFunctionAttributeRector/Fixture/skip_globals_interface.php.inc
@@ -1,0 +1,26 @@
+<?php
+
+namespace Rector\Symfony\Tests\Symfony73\Rector\Class_\GetFunctionsToAsTwigFunctionAttributeRector\Fixture;
+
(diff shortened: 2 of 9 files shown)


TomasVotruba:
LGTM :+1:

PR #948, opened 2026-06-04
[Symfony81] Add new rule for deprecated validator test usages

florianhofsaessC24:
According to https://symfony.com/blog/new-in-symfony-8-1-validator-improvements the validator usage needs to change with Symfony 8.1

Tests using ConstraintValidatorTestCase should call $this->validate(...) (a new helper) instead of $this->validator->validate(...).


TomasVotruba:
Thank you 🙏

PR #959, opened 2026-07-13
[Symfony62] Add new rule for `Command` `configure()` return type

MrYamous:
Based on rules in 6.0 set, this handle `configure()` return type for class extending `Command`


TomasVotruba:
Thanks

PR #960, opened 2026-07-14
[Twig30] Add new set for Twig 30

MrYamous:
Twig 3 is the first one requiring PHP 8.1, which is triggering deprecations if return type is not present


TomasVotruba:
Thanks for kick off :+1:

PR #979, opened 2026-08-01
[composer-based] Bond version-specific rules to composer package constraints

TomasVotruba:
Same shape as rector-phpunit [#754](https://github.com/rectorphp/rector-phpunit/pull/754) / [#747](https://github.com/rectorphp/rector-phpunit/pull/747), applied to Symfony.

## What

A single new set, `config/sets/symfony/composer-based.php`, holding rules bound to the exact Symfony package version their target API was added in. Each rule declares its own constraint:

```php
final class SecurityAttributeToIsGrantedAttributeRector extends AbstractRector implements ComposerPackageConstraintInterface
{
    public function provideComposerPackageConstraint(): ComposerPackageConstraint
    {
        return new ComposerPackageConstraint('symfony/security-http', '>=6.2');
    }
}
```

So the rule runs on any project with that package installed, no matter which Symfony version the upgrade started from - instead of only when the `symfony62` set is picked.

## Effect

A project on `symfony/security-http` 6.4, upgrading from Symfony 5.4, gets the `#[IsGranted]` conversion without running the 6.2 set:

```diff
-#[Security("is_granted('ROLE_ADMIN')")]
+#[IsGranted('ROLE_ADMIN')]
 public function index()
 {
 }
```

And the same rule is skipped on a project still on `symfony/security-http` 5.4, where the attribute does not exist yet.

## Trigger

Symfony has no single package like `phpunit/phpunit` to trigger on, so every package used inside the set registers it, from the lowest version its rules require:

```php
private const array COMPOSER_BASED_TRIGGER_PACKAGES = [
    'symfony/config' => '>=4.2',
    'symfony/process' => '>=4.2',
    'symfony/event-dispatcher' => '>=4.3',
    'symfony/console' => '>=4.4',
    // ...
];
```

## Notes

* Existing sets stay untouched - the rules are registered in both places, and the constraint makes the duplicate harmless.
* `composer.json` is untouched as well. A rule can only be bonded once its package is a dev dependency, otherwise the constraint filters it out and its test silently stops transforming. Rules for `symfony/messenger`, `symfony/twig-bridge`, `symfony/doctrine-bridge` and `symfony/framework-bundle` are therefore left for a follow-up.
* Same for the 7.3 rules of `symfony/console`, `symfony/twig-bundle` and `symfony/security-core`: `rector/rector-src` requires `symfony/console ^6.4.24` and the root requires `symfony/config ^8.1`, so those versions cannot be installed here at all.


TomasVotruba:
First wave, let's ship to allow testing in core.

PR #990, opened 2026-08-02
[composer-based] Bond RedirectToRouteRector to symfony/framework-bundle 2.6

TomasVotruba:
`RedirectToRouteRector` rewrites the redirect helper pair into the shorter route helper:

```diff
 final class SomeController extends AbstractController
 {
     public function someAction()
     {
-        return $this->redirect($this->generateUrl('something', ['id' => $id]));
+        return $this->redirectToRoute('something', ['id' => $id]);
     }
 }
```

`Controller::redirectToRoute()` was added in symfony/framework-bundle 2.6, so the rule now declares that package constraint and joins the composer-based set:

```php
public function provideComposerPackageConstraint(): ComposerPackageConstraint
{
    return new ComposerPackageConstraint('symfony/framework-bundle', '>=2.6');
}
```

A project on an older framework-bundle gets the rule skipped instead of a call to a method that does not exist yet.

## Heads-up on the dependency change

`ComposerPackageConstraintFilter` resolves the constraint against the *installed* packages, so a rule bonded to a package that is not in `require-dev` is filtered out during its own test run and every fixture fails. Keeping the test green therefore requires `symfony/framework-bundle` as a dev dependency.

That does not resolve against the current dev constraints, which sit on three different Symfony lines at once (`dependency-injection ^6.4`, `http-kernel ^7.4`, `config ^8.1`). Working through the majors:

- `^8.1` conflicts with `symfony/console < 8.1`, and `rector/rector-src` pins `symfony/console ^6.4.24` — unresolvable
- `^6.4` needs `http-kernel ^6.4` and `config ^6.1|^7.0`, both below the current root constraints
- `^7.4` resolves, once the packages still pinned at `^6.4` move up

So this PR also moves `dependency-injection`, `routing`, `security-core`, `security-http`, `validator` and `web-link` from `^6.4` to `^7.4`. `http-kernel` was already `^7.4` and `config` stays at `^8.1`.

The full suite passes on the new versions (618 tests, 779 assertions), so nothing in the fixtures appears to depend on the 6.4 APIs. Still, this widens the PR beyond registering one rule — if the `^6.4` pins were deliberate, say so and I will drop the bonding instead.


TomasVotruba:
Closing — keeping the dev dependencies as they are means this rule cannot be bonded.

`ComposerPackageConstraintFilter` resolves `provideComposerPackageConstraint()` against the packages in `vendor/composer/installed.json`. A package that is not installed resolves to `null` and the rule is filtered out, including during its own test run, so all 5 `RedirectToRouteRector` fixtures fail with no change made.

`Controller::redirectToRoute()` is a `symfony/framework-bundle` API, and framework-bundle does not resolve against the current dev constraints at any major:

- `^8.1` conflicts with `symfony/console < 8.1`, while `rector/rector-src` pins `symfony/console ^6.4.24`
- `^7.4` requires `dependency-injection`/`routing` at `^7.4`, currently `^6.4`
- `^6.4` requires `http-kernel ^6.4` and `config ^6.1|^7.0`, currently `^7.4` and `^8.1`

Every rule already in the composer-based set is bonded to a package that happens to be installed, several of them transitively (`console`, `process`, `event-dispatcher`, `http-foundation`, `property-info`, `property-access`). There is no test-side bypass for the filter, and registering the rule in the set without the constraint would break the set's contract that every rule declares its package and version.

`RedirectToRouteRector` stays in `symfony26.php` and `symfony-code-quality.php` as before. Worth revisiting if `symfony/framework-bundle` ever lands in `require-dev` for other reasons.

TomasVotruba:
Superseded by #991, which carries the same bonding without the `composer.json` change. This one could not be reopened — its head commit was gone after the branch delete, and main has since moved with #989.

PR #1018, opened 2026-08-06
[SYMFONY73] add failing set test

JohJohan:
I expect the test to be successful: `Rector\Symfony\Symfony73\Rector\Class_\GetFiltersToAsTwigFilterAttributeRector` and `Rector\Symfony\Symfony73\Rector\Class_\GetFunctionsToAsTwigFunctionAttributeRector` should replace a extension that contains filters and functions right?


TomasVotruba:
You're right :+1:  there was a bug if extension container both, it was skipped.

Fixed in https://github.com/rectorphp/rector-symfony/pull/1027
Use `GetFiltersAndFunctionsToAsTwigAttributeRector` instead :+1:

PR #1030, opened 2026-08-11
chore: Update dispatch build on push to main to rebuild rector-src

samsonasik:
when on push to main (merged to main), it use rector-src and rebuild scoped build.


samsonasik:
@TomasVotruba ready to test

PR #1054, opened 2026-08-25
Add fixture for CommandConfigureToAttributeRector regression

Jean85:
CommandConfigureToAttributeRector  should not run if the promoted value is a non-constant value, since it cannot be included as an attribute argument


TomasVotruba:
Thanks for the report and fixture! Confirmed the rule emitted invalid PHP (`#[AsCommand(description: $this->description)]`). Fixed in #1057 — the rule now skips non-constant `setX()` values and leaves them in `configure()`. Credited you as co-author.

PR #1070, opened 2026-09-23
Fix PhpAstExtractor import removal in composer-based Symfony set

d-mitrofanov-v:
The composer based Symfony set included a rename rule that mapped `Symfony\Component\Translation\Extractor\PhpAstExtractor` to itself. When Rector processed code using that class, it could remove the use import even though the class was still referenced


TomasVotruba:
Can you add validation to the rename itself to throw an exception if class name is the same?

TomasVotruba:
Also, it seems it should be corrected, not removed:

 * Deprecate `PhpExtractor` in favor of `PhpAstExtractor`

d-mitrofanov-v:
Thanks, fixed it. Now it should replace `PhpExtractor` with `PhpAstExtractor`. Also added test for that

TomasVotruba:
No need for test, this is a core rule.

d-mitrofanov-v:
Removed the test

TomasVotruba:
Thank you :+1:

